PR #8, opened 2022-04-04
AE-1217: Distribution monitor fixes

echoAE:
This work modifies the approach for the Compound distribution monitor.  Previously the approach compared the COMP accrued from a previous block with the COMP transferred in the current block.  That comparison was performed as a division, where the previously accrued COMP was in the denominator.  This approach led to many false positives where the previously accrued COMP was zero or close to zero, resulting in percentages that were Infinity or otherwise very high.

The modified approach is to compared the COMP accrued in a single transaction to a fixed threshold (1000 COMP) and create a finding if the COMP accrued exceeds that threshold.


Nomad-AE (approved):
Looks good, very simple and straightforward.

PR #9, opened 2022-04-06
AE-1029 Distribution Agent Bug Fixes

arbitrary-CodeBeholder:


distribution/src/agent.js
```diff
@@ -97,16 +109,26 @@ function provideHandleTransaction(data) {
...
+
       // determine Comptroller.compAccrued() in previous block
       const blockNumber = txEvent.blockNumber;
-      const prevBlockCompAccrued = await comptrollerContract.compAccrued(transferEvent.args.to, { blockTag: blockNumber-1 })
+      const prevBlockCompAccrued = await comptrollerContract.compAccrued(transferEvent.args.to, { blockTag: blockNumber-1 });
       const prevBlockCompAccruedBN = new BigNumber(prevBlockCompAccrued.toString());
+      
+      // if the previous accrual is zero, our heuristic of using the ratio will not work
+      if ( prevBlockCompAccruedBN.isZero() ) {
+        // TODO: potentially check against a maximum "sane" distribution amount of COMP and alert then even if the previous is zero?
```
echoAE, line 129: I think we should put this check in.  I don't know what a maximum "sane" amount is, so I propose selecting a large value and allowing the community to comment on it, if they would like.

Perhaps start at 1,000 COMP, which is in the 100K USD range.

echoAE (approved):

distribution/src/agent.js
arbitrary-CodeBeholder: created https://arbitraryexecution.atlassian.net/browse/AE-1245 to track this feature update

PR #15, opened 2022-04-25
AE-1247: investigate monitoring large positions that may be unhealthy close to liquidation

Nomad-AE:
Initial implementation of a compound liquidation monitor.


liquidation-monitor/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# Final stage: install production dependencies
+FROM node:12-alpine
+ENV NODE_ENV=production
+# Uncomment the following line to enable agent logging
+# LABEL "network.forta.settings.agent-logs.enable"="true"
+WORKDIR /app
+# if using obfuscated code from build stage:
+# COPY --from=builder /app/dist ./src
+# else if using unobfuscated code:
+COPY ./src ./src
```
echoAE, line 17:
```suggestion
COPY ./src ./src
COPY ./abi ./abi
```

liquidation-monitor/README.md
```diff
@@ -0,0 +1,21 @@
+# Compound Liquidation Monitor
+
+## Description
+
+This agent detects when an account on Compound is able to be liquidated.
```
echoAE, line 5:
```suggestion
This bot detects when an account on Compound is able to be liquidated.
```

liquidation-monitor/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Optimizations
+
+- Although there are ~400k accounts using Compound, limit the search to 1 wei ETH borrowed it will reduce the tracked accounts down to just over 9000. A nearly 98% reduction in assets being tracked.
```
echoAE, line 9:
```suggestion
- Although there are ~400k accounts using Compound, limiting the search to 1 wei ETH borrowed reduces the tracked accounts down to just over 9000. A nearly 98% reduction in assets being tracked.
```

liquidation-monitor/README.md
```diff
@@ -0,0 +1,21 @@
...
+- Lowering the maxHealth setting will further reduce this list.
+- Initial account information can be obtained from Compound API
```
echoAE, line 11:
```suggestion
- Initial account information can be obtained from Compound API.
```

liquidation-monitor/abi/Comptroller.json
```diff
@@ -0,0 +1,2299 @@
+{
```
echoAE, line 1: Can you replace all of the tabs with 2 spaces.

liquidation-monitor/abi/FeedRegistry.json
```diff
@@ -0,0 +1,511 @@
+{
```
echoAE, line 1: Can you replace each of the tabs with 2 spaces?

liquidation-monitor/abi/OneInch.json
```diff
@@ -0,0 +1,336 @@
+{
```
echoAE, line 1: Can you replace each of the tabs with 2 spaces?

liquidation-monitor/abi/Oracle.json
```diff
@@ -0,0 +1,986 @@
+{
```
echoAE, line 1: Can you replace each of the tabs with 2 spaces?

liquidation-monitor/abi/cErc20.json
```diff
@@ -0,0 +1,1466 @@
+{
```
echoAE, line 1: Can you replace each of the tabs with 2 spaces?

liquidation-monitor/compound-openrpc.json
```diff
@@ -0,0 +1,179 @@
+{
```
echoAE, line 1: Can you replace each of the tabs with 2 spaces?

liquidation-monitor/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "forta-agent-starter",
```
echoAE, line 2:
```suggestion
  "name": "forta-compound-finance-liquidation-monitor",
```

liquidation-monitor/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "version": "0.0.1",
+  "description": "Forta Agent Javascript starter project",
```
echoAE, line 4:
```suggestion
  "description": "Forta Bot that monitors Compound Finance for positions that can be liquidated",
```

liquidation-monitor/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "push": "forta-agent push",
+    "disable": "forta-agent disable",
+    "enable": "forta-agent enable",
+    "keyfile": "forta-agent keyfile",
+    "test": "jest"
+  },
+  "dependencies": {
+    "bignumber.js": "^9.0.2",
+    "forta-agent": "^0.1.3",
+    "node-fetch-commonjs": "^3.1.1"
```
echoAE, line 26: Were we still hoping to use `axios` or stick with `fetch`?

liquidation-monitor/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+    .then((responseData) => responseData);
+
+  const responseData = await callPromise.then(
+    (data) => data,
+    (error) => {
+      console.error('Failed to fetch accounts due to error: ', error);
+    },
+  );
+  return responseData;
+}
```
echoAE, line 40: Replace with `axios`?

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
...
+const BigNumber = require('bignumber.js');
+
+// To-do: Replace node-fetch with axios and remove TS
+const config = require('../agent-config.json');
+const { getAbi, ts, callAPI } = require('./utils');
+
+// Stores information about each account
+const initializeData = {};
+
+ts('Starting bot');
```
echoAE, line 17: Remove these `ts` calls.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
+const {
+  Finding,
+  FindingSeverity,
+  FindingType,
+  ethers,
+  getEthersProvider,
+} = require('forta-agent');
+const BigNumber = require('bignumber.js');
+
+// To-do: Replace node-fetch with axios and remove TS
```
echoAE, line 10: Do what the `To-do` states and remove comment.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
...
+    if (data.supply[tokenAddress] === undefined) {
+      data.supply[tokenAddress] = {};
+    }
+    /* eslint-enable no-param-reassign */
+
+    ts(
+      `Now tracking ${data.tokens[tokenAddress].symbol
+      } with ${data.tokens[tokenAddress].tokenDecimals
+      } decimals at address ${tokenAddress}`,
+    );
```
echoAE, line 92:
```suggestion
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
...
+    // cETH does not have an underlying contract, so peg it to wETH instead
+    if (tokenAddress === '0x4ddc2d193948926d02f9b1fe9e1daa0718270ed5') {
+      data.tokens[tokenAddress].underlying = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2';
+    } else {
+      data.tokens[tokenAddress].underlying = await cContract.underlying();
+    }
+    const exchangeRate = await cContract.exchangeRateStored();
+    data.tokens[tokenAddress].cTokenDecimals = await cContract.decimals();
+    // Token to cToken is approximately 0.02 * 10**(10 + tokenDecimals)
+    // So the trick to get token decimals is exchangeRate.lenth - 9
```
echoAE, line 67:
```suggestion
    // So the trick to get token decimals is exchangeRate.length - 9
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
...
+      feedRegistryABI,
+      data.provider,
+    );
+    // ChainLink calls don't work expected.
+    // https://docs.chain.link/docs/feed-registry/
+    // const ethDenom = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
+    // const cUSDDenom = '0x39aa39c021dfbae8fac545936693ac917d5e7563';
+    // const test = await data.feedRegistryContract.latestRoundData(cUSDDenom, ethDenom);
+    // Errors with:too many arguments: passed to contract
+    // (count = 2, expectedCount = 0, code = UNEXPECTED_ARGUMENT, version = contracts / 5.6.0)
```
echoAE, line 155:
```suggestion
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
...
+    const initialRequest = buildJsonRequest(
+      maximumHealth,
+      minimumBorrowInETH,
+      1,
+      1,
+    );
+    const initialResults = await callAPI(apiURL, initialRequest);
+    const totalEntries = initialResults.pagination_summary.total_entries;
+    ts(`Total Entries ${totalEntries}`);
+    ts(`maxTrackedAccounts ${maxTrackedAccounts}`);
```
echoAE, line 191:
```suggestion
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
...
+        maximumHealth,
+        minimumBorrowInETH,
+        page,
+        100, // 100 results per page was optimal
+      );
+      const apiResults = await callAPI(apiURL, currentRequest);
+      apiResults.accounts.forEach((account) => {
+        accounts.push(account);
+      });
+      ts(`Imported batch ${page}00 accounts`);
```
echoAE, line 211:
```suggestion
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
...
+    }));
+    ts(`Tracking ${accounts.length} accounts`);
+    // #endregion
+
+    // #region Parse Compound data into new objects
+    // Get a unique list of token addresses
+    ts('Processing tokens now');
```
echoAE, line 218:
```suggestion
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,522 @@
...
+      });
+    });
+
+    // Initialize token objects first
+    await Promise.all(Array.from(tokenSet).map(async (token) => {
+      await verifyToken(data, token);
+    }));
+
+    // Loop through found accounts
+    ts('Updating user balances');
```
echoAE, line 232:
```suggestion
```

liquidation-monitor/abi/Comptroller.json
Nomad-AE: Can do!

liquidation-monitor/package.json
Nomad-AE: Fetch works well, I will leave it as is.

liquidation-monitor/src/utils.js
Nomad-AE: I will keep fetch and remove my comments

liquidation-monitor/src/agent.js
Nomad-AE: I left it verbose to to be able to see what's going on at each step. To mute the `ts` calls, I can comment out the `console.log` line in the function to be able to turn troubleshooting and debug on an off. I think leaving some of the logging would be helpful for future iterations. What do you think?

liquidation-monitor/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+      return response.json();
+    })
+    .then((responseData) => responseData);
+
+  const responseData = await callPromise.then(
+    (data) => data,
+    (error) => {
+      console.error('Failed to fetch accounts due to error: ', error);
+    },
+  );
```
echoAE, line 38: For readability and consistency, we prefer to not mix `then` with `await`.

By convention, we use `await` throughout unless there is a compelling reason to do otherwise.

The following code _should_ be equivalent to the original (this suggestion is untested).
```suggestion
  const response = await fetch(url, {
    method: 'POST',
    body: JSON.stringify(jsonRequest),
  });

  if (!response.ok) {
    console.error('Got non-2XX response from API server.');
  }
  
  let responseData;
  try {
    responseData = await response.json();
  } catch (error) {
    console.error('Failed to fetch accounts due to error: ', error);
  }
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+    /* eslint-enable no-param-reassign */
+    // #endregion
+
+    // #region Get initial accounts from Compound API
+    // To-do: Can be bootstrapped with ~30 eth_getLogs calls,  500k blocks per call, using filter:
+    // Borrow (address borrower, uint256 borrowAmount, uint256 accountBorrows, uint256 totalBorrows)
+    // topic0: 0x13ed6866d4e1ee6da46f845c46d7e54120883d75c5ea9a2dacc1c4ca8984ab80
+
+    // Simple bootstrap with Compound for now
+    const apiURL = 'https://api.compound.finance/api/v2/account';
```
echoAE, line 148: This constant isn't used anywhere other than the `callAPI` function, which is imported from another file.  This should be moved to that file and removed as an argument to the `callAPI` function.  The `callAPI` function could probably also be renamed `callCompoundAPI` or something similar, as the request and response formats appear to be specific to that particular API.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+
+    // Query each page and add accounts. Starting at 1 and including maxPages
+    const accounts = [];
+    const pages = [...Array(maxPages)].map((_, i) => 1 + i);
+    await Promise.all(pages.map(async (page) => {
+      const currentRequest = buildJsonRequest(
+        maximumHealth,
+        minimumBorrowInETH,
+        page,
+        100, // 100 results per page was optimal
```
echoAE, line 185: Could you provide a little more detail about how it was `optimal`?  Was it empirically determined based on the time required to perform all requests and parse all responses?

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+
+    // #region Parse Compound data into new objects
+    // Get a unique list of token addresses
+    ts('Processing tokens now');
+    const tokenSet = new Set();
+    accounts.forEach((account) => {
+      account.tokens.forEach(async (token) => {
+        tokenSet.add(token.address);
+      });
+    });
```
echoAE, line 202: I don't know enough about the implementation of `Set`, but have you considered constructing the Array and then passing it to the `Set` constructor instead?  I'm guessing that any performance benefit will rely upon how many duplicates you expect (the number of accounts and tokens is probably so low as to be in the noise of any performance difference).

@ae-cloin may have some insight, but this is my first crack at it:
```suggestion
    const tokens = accounts.map((account) => account.tokens.map((token) => token.address)).flat();
    const tokenSet = new Set(tokens);
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+  if (data.tokens[tokenAddress] === undefined) {
+    data.tokens[tokenAddress] = {};
+    data.tokens[tokenAddress].contract = new ethers.Contract(
+      tokenAddress,
+      data.cTokenABI,
+      data.provider,
+    );
+    const cContract = data.tokens[tokenAddress].contract;
+    data.tokens[tokenAddress].symbol = await cContract.symbol();
+    // cETH does not have an underlying contract, so peg it to wETH instead
```
echoAE, line 54: Nit: Insert an empty line to break up this large block of text.
```suggestion

    // cETH does not have an underlying contract, so peg it to wETH instead
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+      }
+      // Add found health
+      data.accounts[account.address].health = BigNumber(account.health.value);
+      // Loop through tokens and update balances.
+      account.tokens.forEach((token) => {
+        // Process borrows as 'token'
+        /* eslint-disable eqeqeq */
+        if (
+          token.borrow_balance_underlying !== undefined
+          && token.borrow_balance_underlying.value != 0
```
echoAE, line 225: Is this intended to be `!=` or `!==`?
```suggestion
          && token.borrow_balance_underlying.value !== 0
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+          ts(
+            `User ${[account.address]
+            } borrowed ${Math.round(token.borrow_balance_underlying.value)
+            } ${data.tokens[token.address].symbol}`,
+          );
+        }
+        // Process supplies as 'cTokens'
+        if (
+          token.supply_balance_underlying !== undefined
+          && token.supply_balance_underlying.value != 0
```
echoAE, line 240: Same question as the previous one.
```suggestion
          && token.supply_balance_underlying.value !== 0
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+      `Currently tracking ${Object.keys(data.accounts).length
+      } accounts and adding ${data.newAccounts.length
+      } account(s) this block. ${data.totalNewAccounts
+      } total accounts added since start.`,
+    );
+    data.newAccounts.forEach((newAccount) => {
+      const account = newAccount.toLowerCase();
+      // Initialize account. New accounts will get updated in the block section
+      /* eslint-disable no-param-reassign */
+      data.accounts[account] = {};
```
echoAE, line 303: Nit: You can eliminate the extra variable.
```suggestion
      // Initialize account. New accounts will get updated in the block section
      /* eslint-disable no-param-reassign */
      data.accounts[newAccount.toLowerCase()] = {};
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+      /* eslint-enable no-param-reassign */
+    });
+    /* eslint-disable no-param-reassign */
```
echoAE, line 306: These eslint commands can probably be eliminated (no need to toggle on/off the same setting).
```suggestion
    });
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+    // #endregion
+
+    // #region Update Balances on zero health accounts
+    const filteredAccounts = [];
+    Object.keys(data.accounts).forEach((currentAccount) => {
+      if (data.accounts[currentAccount].health == null
+        || data.accounts[currentAccount].health === 0) {
+        filteredAccounts.push(currentAccount);
+        // Zero account balances
+        /* eslint-disable no-param-reassign */
```
echoAE, line 319: Actually, given how often this eslint setting is toggle on/off, perhaps the setting should just be left disabled for the entire function.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+    // Grab the assets in first, and make sure they are initialized
+    ts('Getting assetsIn list for new accounts');
+    const tokenSet = new Set();
+    await Promise.all(filteredAccounts.map(async (currentAccount) => {
+      let assetsIn = await comptrollerContract.getAssetsIn(currentAccount);
+      assetsIn = assetsIn.map((asset) => asset.toLowerCase());
+      assetsIn.forEach((asset) => { tokenSet.add(asset); });
+      /* eslint-disable no-param-reassign */
+      data.accounts[currentAccount].assetsIn = assetsIn;
+      /* eslint-enable no-param-reassign */
```
echoAE, line 339: Assuming the eslint disable/enable comments are applied to the whole function.

This is just an alternative suggestion, potentially more readable?
```suggestion
      const assetsIn = await comptrollerContract.getAssetsIn(currentAccount);
      data.accounts[currentAccount].assetsIn = assetsIn.map((asset) => {
        const address = asset.toLowerCase();
        tokenSet.add(address);
        return address;
      });
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+      const market = await comptrollerContract.markets(currentToken);
+      data.tokens[currentToken].collateralMult = BigNumber(market[1].toString())
+        .dividedBy(BigNumber(10).pow(18));
+      /* eslint-enable no-param-reassign */
+      ts(
+        `Updated price of ${data.tokens[currentToken].symbol
+        } is ${data.tokens[currentToken].price.toString()
+        } ETH with a collateral factor of ${data.tokens[currentToken].collateralMult}`,
+      );
+    }));
```
echoAE, line 404: You should be able to use `entries` in place of `keys` to make the code more readable.
```suggestion
    await Promise.all(Object.entries(data.tokens).map(async (currentToken, tokenData) => {
      const price = await oneInchContract.getRateToEth(tokenData.underlying, 0);
      
      // Adjust for native decimals
      const oneInchMult = BigNumber(10).pow(36 - tokenData.tokenDecimals);
      tokenData.price = BigNumber(price.toString()).dividedBy(oneInchMult);
      
      // Update the Collateral Factor
      const market = await comptrollerContract.markets(currentToken);
      tokenData.collateralMult = BigNumber(market[1].toString())
        .dividedBy(BigNumber(10).pow(18));
    }));
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+    // Return zero findings
+    return [];
+  };
+}
+
+function provideHandleBlock(data) {
+  return async function handleBlock(blockEvent) {
+    ts(`Starting block ${blockEvent.blockNumber}`);
+    const findings = [];
+    const { comptrollerContract, oneInchContract } = data;
```
echoAE, line 290: If you destructure `data` to also pull several other frequently used objects (e.g. `tokens`, `accounts`, `borrow`, etc.), it may improve the readability of this function (at least you'll be lopping off the `data.` from each time you use them):
```suggestion
    const {
      comptrollerContract,
      oneInchContract,
      accounts,
      tokens,
      borrow,
      supply,
    } = data;
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+    // #region Check for low health accounts currently health <1.03
+    const lowHealthAccounts = [];
+    Object.keys(data.accounts).forEach((currentAccount) => {
+      if (data.accounts[currentAccount].health.isLessThan(1.03)) {
+        lowHealthAccounts.push(currentAccount);
+      }
+    });
+    ts(`${lowHealthAccounts.length} low health accounts detected`);
+    await Promise.all(lowHealthAccounts.map(async (currentAccount) => {
+      const liquidity = await data.comptrollerContract.getAccountLiquidity(currentAccount);
```
echoAE, line 450:
```suggestion
      const liquidity = await comptrollerContract.getAccountLiquidity(currentAccount);
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+      let borrowBalance = BigNumber(0);
+      let supplyBalance = BigNumber(0);
+      Object.keys(data.tokens).forEach((token) => {
+        if (data.supply[token][account]) {
+          // Supply balances are stored in cTokens and need to be multiplied by the
+          //   exchange rate and reduced by the collateral factor
+          supplyBalance = supplyBalance.plus(data.supply[token][account]
+            .multipliedBy(data.tokens[token].exchangeRateMult)
+            .multipliedBy(data.tokens[token].price)
+            .multipliedBy(data.tokens[token].collateralMult));
```
echoAE, line 419: This comment does not appear to match the code it describes: `reduced by the collateral factor` implies either a subtraction or division, but this uses a multiplication.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+        data.accounts[account].health = supplyBalance.dividedBy(borrowBalance);
+        /* eslint-enable no-param-reassign */
+      }
+    });
+    // #endregion
+
+    // #region Check for low health accounts currently health <1.03
+    const lowHealthAccounts = [];
+    Object.keys(data.accounts).forEach((currentAccount) => {
+      if (data.accounts[currentAccount].health.isLessThan(1.03)) {
```
echoAE, line 444: Make the health threshold a parameter that is specified in the `agent-config.json`.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+        //   } and cash reserve of +$${positiveUSD.dp(2)} USD`,
+        // );
+      }
+      if (shortfallUSD.gt(0)) {
+        ts(
+          `Negative liquidity on ${currentAccount} with health of ${data.accounts[currentAccount].health.dp(3)
+          } and shortfall of +$${shortfallUSD.dp(2)} USD`,
+        );
+      }
+      // Lower health means that less money that can be retrieved.
```
echoAE, line 468: Is this comment accurate?

Isn't health the ratio between amount supplied and amount borrowed?  So a low ratio doesn't indicate anything about the amount that can be retrieved.

liquidation-monitor/src/agent.spec.js
```diff
@@ -0,0 +1,72 @@
+const {
```
echoAE, line 1: This looks like the starter Bot test script example.  Need to create tests for this Bot.

echoAE (changes requested):
I've offered some (hopefully) simplifying suggestions, as well as some additional checks that need to be added to the `filterLog` calls.  The toggling of eslint warnings on and off throughout the script should be reduced to a single `eslint-disable` for the script. 
 We are disabling the `no-param-reassign` warning in every place that it could possibly occur, which means we should just consolidate and do it once rather than peppering it amongst the functional code).

The troubleshooting and debugging mechanisms (e.g. `ts()`) should be removed, as well as any unused code and todo comments.

We need to work on creating test scripts for this.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+    const exitMarketEvents = txEvent.filterLog(exitMarketString);
+    exitMarketEvents.forEach((exitEvent) => {
+      ts(`${exitEvent.args.account} exited from ${exitEvent.args.account} `);
+      data.newAccounts.push(exitEvent.args.account);
+    });
+    const borrowEvents = txEvent.filterLog(borrowString);
+    borrowEvents.forEach((borrowEvent) => {
+      ts(`${borrowEvent.args.borrower} borrowed ${borrowEvent.args.borrowAmount} `);
+      data.newAccounts.push(borrowEvent.args.borrower);
+    });
```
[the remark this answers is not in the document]
Nomad-AE: `MarketExited` can definitely be filtered by the Comptroller address. The `Borrow` address was intentionally omitted to ensure that I would pick up newly listed Compound tokens. There will be some collisions with other token Events, but in those cases, the detected account will be added to the `newAccount` list. At the beginning of the next block, the `newAccount` will be scanned for Compound assets, and if none are found, the account will be discarded. 

This approach uses less calls than calling `Comptroller.getAllMarkets()` every block.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+          data.borrow[currentToken][currentAccount] = BigNumber(snapshot[2].toString())
+            .dividedBy(data.tokens[currentToken].tokenDecimalsMult);
+          ts(
+            `User ${currentAccount
+            } borrowed ${Math.round(data.borrow[currentToken][currentAccount])
+            } ${data.tokens[currentToken].symbol}`,
+          );
+        }
+        /* eslint-enable no-param-reassign */
+        return null;
```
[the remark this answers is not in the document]
Nomad-AE: Safe assumption. I agree. And there are no divide by zero issues. Much cleaner. Thank you.

liquidation-monitor/src/agent.js
Nomad-AE: I agree, I will check for variable collisions before I make this change.

liquidation-monitor/src/agent.js
Nomad-AE: Agreed. Actually the destructure will help in this regard

liquidation-monitor/src/agent.js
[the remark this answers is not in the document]
echoAE: This sounds like an approach that will work.

Nomad-AE:
Will leave the `ts` function in until I finish implementing tests.

liquidation-monitor/src/agent.js
Nomad-AE: Corrected in commit 79512957480ad93f994c11aa269380233ad83787

liquidation-monitor/src/agent.js
Nomad-AE: Corrected in commit 79512957480ad93f994c11aa269380233ad83787

liquidation-monitor/src/agent.js
Nomad-AE: Removed comments

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,500 @@
...
+      const jsonRequest = {
+        addresses: [], // returns all accounts if empty or not included
+        block_number: 0, // returns latest if given 0
+        max_health: { value: maxHealth },
+        min_borrow_value_in_eth: { value: minBorrow },
+        page_number: pageNumber,
+        page_size: pageSize,
+      };
+      return jsonRequest;
+    }
```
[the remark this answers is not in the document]
Nomad-AE: Corrected in commit bd67d88efe6427a18d45e91b638aebe719f76827

liquidation-monitor/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+  const today = new Date();
+  const hours = String(today.getHours()).padStart(2, '0');
+  const minutes = String(today.getMinutes()).padStart(2, '0');
+  const seconds = String(today.getSeconds()).padStart(2, '0');
+  const time = `${hours}:${minutes}:${seconds}`;
+  console.log('TS -', time, input);
+}
+
+// To-do: Replace fetch with axios
+const fetch = require('node-fetch-commonjs');
```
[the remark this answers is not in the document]
Nomad-AE: Corrected in commit bd67d88efe6427a18d45e91b638aebe719f76827

liquidation-monitor/src/agent.js
Nomad-AE: Corrected in commit bd67d88efe6427a18d45e91b638aebe719f76827

liquidation-monitor/src/agent.js
Nomad-AE: Removed comment

liquidation-monitor/src/agent.js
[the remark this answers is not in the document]
Nomad-AE: `MarketExited` is now filtered by comptroller address per commit 4f4872a475d3c6a5a162e96c76b34ac703417ec8

liquidation-monitor/src/agent.js
Nomad-AE: Less toggling of linting on commit da9b6aa2a0142c6d6330c6f3e78b4d103959dad8

Nomad-AE:
I've addressed everything that you @echoAE mentioned. If it looks good, I'll move on to the testing.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+      const price = await oneInchContract.getRateToEth(tokens[currentToken].underlying, 0);
+      const oneInchMult = BigNumber(10).pow(36 - tokens[currentToken].tokenDecimals);
+
+      // Adjust for native decimals
+      tokens[currentToken].price = BigNumber(price.toString()).dividedBy(oneInchMult);
+
+      // Update the Collateral Factor
+      const market = await comptrollerContract.markets(currentToken);
+      tokens[currentToken].collateralMult = BigNumber(market[1].toString())
+        .dividedBy(BigNumber(10).pow(18));
```
arbitrary-sparrow, line 317: Add a comment explaining that the collateralFactorMantissa (`market[1]`) is scaled by `1e18`

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+    );
+    /* eslint-enable no-param-reassign */
+    // #endregion
+
+    // #region Get initial accounts from Compound API
+    // Find total number of results with the first request
+    const initialRequest = buildJsonRequest(maximumHealth, minimumBorrowInETH, 1, 1);
+    const initialResults = await callCompoundAPI(initialRequest);
+    const totalEntries = initialResults.pagination_summary.total_entries;
+
```
ae-cloin, line 148: There's a bit of a race condition here. If you grab total pages and then iterate assuming total pages never changes, and transactions are made mid page request loop, things get a little squirrely.. 
Unfortunately, this form of pagination is just messy if not done very carefully, this is not on you. Hopefully new transactions get added to the end of the list, otherwise you may wind up with duplicates.

The only vaguely ok way to do this is just to loop over the stuff in L157->L162 until the callCompoundAPI request fails due to an out-of-range, and hope you got a coherent snapshot. You can then get/use the length of your foundAccounts array if you need a max length from there.

arbitrary-sparrow (changes requested):

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+    // Initialize token objects first
+    await Promise.all(filteredTokens.map(async (token) => {
+      await verifyToken(data, token);
+    }));
+
+    // Loop through found accounts
+    foundAccounts.forEach((account) => {
+      // add to tracked accounts
+      /* eslint-disable no-param-reassign */
```
arbitrary-sparrow, line 182: It seems like this gets disabled/enabled often, so it should probably be disabled for the entire file or via the `.eslintrc`

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+    // #endregion
+
+    // Return zero findings
+    return [];
+  };
+}
+
+function provideHandleBlock(data) {
+  // eslint-disable-next-line no-unused-vars
+  return async function handleBlock(blockEvent) {
```
arbitrary-sparrow, line 235:
```suggestion
  return async function handleBlock() {
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+    // #region Get initial accounts from Compound API
+    // Find total number of results with the first request
+    const initialRequest = buildJsonRequest(maximumHealth, minimumBorrowInETH, 1, 1);
+    const initialResults = await callCompoundAPI(initialRequest);
+    const totalEntries = initialResults.pagination_summary.total_entries;
+
+    // Determine number of pages needed to query. Results vs config limit.
+    const maxEntries = Math.min(maxTrackedAccounts, totalEntries);
+    const maxPages = Math.ceil(maxEntries / 100);
+
```
ae-cloin, line 152: This is potentially messy. Depending on how they've implemented result ordering, (if they sort newest -> oldest, so new transactions show up in earlier pages and push other data back) you can potentially get duplicate data, maxPages/entries might be stale by the time you've walked the result, etc

Not sure if we can really do anything, it's an API problem, but it's worth mentioning. Results from this may wind up looking a bit strange with high transaction count traffic.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+      }));
+    }));
+    // #endregion
+
+    // #region Update all token prices via 1inch for now.
+    // Mapping through all tokens
+    // Note: Object.entries does not work here because the nested objects cannot be retrieved.
+    await Promise.all(Object.keys(tokens).map(async (currentToken) => {
+      const price = await oneInchContract.getRateToEth(tokens[currentToken].underlying, 0);
+      const oneInchMult = BigNumber(10).pow(36 - tokens[currentToken].tokenDecimals);
```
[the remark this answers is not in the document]
Nomad-AE: Honestly, it was trial and error to arrive at this, I'm guessing that the standard token has 18 decimals and they also wanted the price to have an 18 decimal precision. So, 18+18 gives you the 36 decimals in the multiplier. Sounds good, but what if they token only uses 6 decimals, then they use the remaining 30 decimals for pricing. I'll add a comment to clarify this point.

liquidation-monitor/src/agent.js
Nomad-AE: Correct, I do not use the blockEvent data, I wasn't sure if it would cause an error if I removed the parameter here.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+const BigNumber = require('bignumber.js');
+const config = require('../agent-config.json');
+const {
+  getAbi, callCompoundAPI, buildJsonRequest,
+} = require('./utils');
+
+// Stores information about each account
+const initializeData = {};
+
+// #region Global functions
```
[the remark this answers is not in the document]
Nomad-AE: VSC specific. I'll remove them.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+      tokens[tokenAddress].underlying = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2';
+    } else {
+      tokens[tokenAddress].underlying = await cContract.underlying();
+    }
+    const exchangeRate = await cContract.exchangeRateStored();
+    tokens[tokenAddress].cTokenDecimals = await cContract.decimals();
+
+    // Token to cToken is approximately 0.02 * 10**(10 + tokenDecimals)
+    // So the trick to get token decimals is exchangeRate.length - 9
+    tokens[tokenAddress].tokenDecimals = exchangeRate.toString().length - 9;
```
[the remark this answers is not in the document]
Nomad-AE: Much cleaner, thank you.

liquidation-monitor/src/agent.js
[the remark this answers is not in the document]
Nomad-AE: Changed in commit 9e1c0fedc3c471f12d898f61276b5bd4f68021d1

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,404 @@
...
+    tokens[tokenAddress].symbol = await cContract.symbol();
+
+    // cETH does not have an underlying contract, so peg it to wETH instead
+    if (tokenAddress === '0x4ddc2d193948926d02f9b1fe9e1daa0718270ed5') {
+      tokens[tokenAddress].underlying = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2';
+    } else {
+      tokens[tokenAddress].underlying = await cContract.underlying();
+    }
+    const exchangeRate = await cContract.exchangeRateStored();
+    tokens[tokenAddress].cTokenDecimals = await cContract.decimals();
```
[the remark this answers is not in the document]
Nomad-AE: Resolved in commit 4d08167f9589157369e61c05a41a12b9af9e3385

liquidation-monitor/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+    min_borrow_value_in_eth: { value: minBorrow },
+    page_number: pageNumber,
+    page_size: pageSize,
+  };
+  return jsonRequest;
+}
+
+async function callCompoundAPI(jsonRequest) {
+  const apiURL = 'https://api.compound.finance/api/v2/account';
+  const response = await fetch(apiURL, {
```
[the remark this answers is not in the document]
Nomad-AE: Resolved in commit b2370153ed9e8e58aca433ed5d6b5574358d2263

liquidation-monitor/src/agent.js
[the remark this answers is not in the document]
Nomad-AE: Comments  added.

liquidation-monitor/src/agent.js
Nomad-AE: disabled it in the `.eslintrc` as recommended

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,383 @@
...
+    tokens[tokenAddress].tokenDecimalsMult = BigNumber(10).pow(tokens[tokenAddress].tokenDecimals);
+    tokens[tokenAddress].cTokenDecimalsMult = BigNumber(10)
+      .pow(tokens[tokenAddress].cTokenDecimals);
+    tokens[tokenAddress].exchangeRate = BigNumber(exchangeRate.toString());
+
+    // Adjusting the multiplier for easier use later.
+    tokens[tokenAddress].exchangeRateMult = BigNumber(exchangeRate.toString())
+      .dividedBy(tokens[tokenAddress].tokenDecimalsMult)
+      .dividedBy(tokens[tokenAddress].cTokenDecimalsMult)
+      .dividedBy(100); // Not sure where this 100 comes from I didn't see it in the docs
```
echoAE, line 86: It looks like the value returned from the `exchangeRateStored` method (and the internal `exchangeRateStoredInternal` method) is scaled by 1e18:
https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/CToken.sol#L326
https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/CToken.sol#L337


Note: Be certain not to use ethers.js BigNumber, as the division will cause a loss of precision in the result.  Only use bignumber.js BigNumber, as it will retain precision through the division.
```suggestion
const exchangeRateDecimals = 18;
const multiplier = new BigNumber(10).pow(exchangeRateDecimals);
tokens[tokenAddress].exchangeRateMult = BigNumber(exchangeRate.toString()).div(multiplier);
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,383 @@
...
+    tokens[tokenAddress].symbol = await cContract.symbol();
+
+    // cETH does not have an underlying contract, so link it to wETH instead
+    if (tokenAddress === '0x4ddc2d193948926d02f9b1fe9e1daa0718270ed5') {
+      tokens[tokenAddress].underlying = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2';
+    } else {
+      tokens[tokenAddress].underlying = await cContract.underlying();
+    }
+    const exchangeRate = await cContract.exchangeRateStored();
+    tokens[tokenAddress].cTokenDecimals = BigNumber(
```
echoAE, line 64: Although JavaScript will run this, the constructor for BigNumber should use the `new` keyword:
```suggestion
    tokens[tokenAddress].cTokenDecimals = new BigNumber(
```

This should be done anywhere the BigNumber constructor is used.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,383 @@
...
+
+    // Determine number of pages needed to query.
+    const maxPages = Math.ceil(totalEntries / 100);
+
+    // Query each page and add accounts. Starting at 1 and including maxPages
+    const foundAccounts = [];
+    // Shorthand Range() function. ( Ex: 5 => [1,2,3,4,5] )
+    const pages = [...Array(maxPages)].map((_, i) => 1 + i);
+    await Promise.all(pages.map(async (page) => {
+      const currentRequest = buildJsonRequest(maximumHealth, minimumBorrowInETH, page, 100);
```
echoAE, line 149:
```suggestion
    const pageIncrement = 100;

    // Determine number of pages needed to query.
    const maxPages = Math.ceil(totalEntries / pageIncrement);

    // Query each page and add accounts. Starting at 1 and including maxPages
    const foundAccounts = [];
    // Shorthand Range() function. ( Ex: 5 => [1,2,3,4,5] )
    const pages = [...Array(maxPages)].map((_, i) => 1 + i);
    await Promise.all(pages.map(async (page) => {
      const currentRequest = buildJsonRequest(maximumHealth, minimumBorrowInETH, page, pageIncrement);
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,383 @@
...
+    }));
+
+    // Update all token prices via 1inch
+    // Note: Object.entries does not work here because the nested objects cannot be retrieved.
+    await Promise.all(Object.keys(tokens).map(async (currentToken) => {
+      const price = await oneInchContract.getRateToEth(tokens[currentToken].underlying, 0);
+
+      // 1inch's getRateToEth uses 36 decimal places. The space is shared with the token's decimal.
+      //   So (36 - tokenDecimal) will yield the actual decimal for the rate.
+      const oneInchMult = BigNumber(10).pow(36 - tokens[currentToken].tokenDecimals);
```
echoAE, line 290: Based on the complexity of the 1inch contract math and attempting to compensate for it here, it would be better to select a simpler smart contract oracle.  Options include Chainlink Aggregators, as we used for Aave:
https://github.com/arbitraryexecution/forta-aave-suite/blob/main/compare-oracle-to-fallback/src/agent.js

Or Uniswap, which we used for Gamma (nee Visor): see our Confluence page on how to get the price ratio

liquidation-monitor/src/agent.spec.js
Nomad-AE: Tests completed.

liquidation-monitor/src/agent.js
Nomad-AE: “The current exchange rate as an unsigned integer, scaled by 1 * 10^(18 - 8 + Underlying Token Decimals)” - https://compound.finance/docs/ctokens#exchange-rate

The 8 mentioned here is the cToken Decimals. The exchange rate is not a static 1e18. Example, these are the exchange rates for a few:
225975924421044 for USDC ( uses 6 decimals)
20204487858136665 for wBTC ( uses 8 decimals)
202562285509376818212946300 for UNI ( uses 18 decimals )

So the exchange rate / cToken decimals / token decimals yields the actual rate I need.
Given: 1 cUSDC = 0.02259 USDC
For USDC: 225975924421044 / 10^8 / 10^6  = 2.259  This is still off by 100, so I divided by 100 to get the final multiplier of 0.02259 . 

Also “When a market is launched, the cToken exchange rate (how much ETH one cETH is worth) begins at 0.020000 — and increases at a rate equal to the compounding market interest rate.” - https://compound.finance/docs/ctokens#introduction

This might explain the 100 factor to get exchange rate of 2 to 0.02

liquidation-monitor/src/agent.js
Nomad-AE: Corrected on commit 68d8ca5da630a046bf35cbf095720fa39beccb37

liquidation-monitor/src/agent.js
Nomad-AE: I had issues with my implementation of Chainlink so I removed it in an earlier commit. I don't see this any more complex than the `exchangeRateStored` math in the cTokens, but I can switch it to UNI. I felt that 1inch was more robust because it's price feed is pulled form UNI v1, v2, v3, Chainlink, Sushi and more.

liquidation-monitor/src/agent.js
Nomad-AE: Actually, this part explains it the best. "1 * 10^(18 - 8 + Underlying Token Decimals)"
10^(10 + Underlying Token Decimals). I'll adjust my code to this.

Nomad-AE: Simplified and added more notes in commit 12d462d1b7a29c7db85c9619d9875740f6a01f13

liquidation-monitor/src/agent.js
Nomad-AE: Clarified the 1inch scaling (with source) in commit b1fdb3a86d5ca1d9e57500c9c1a43f77d2d9b138

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,389 @@
...
+    tokens[tokenAddress].cTokenDecimalsMult = new BigNumber(10)
+      .pow(tokens[tokenAddress].cTokenDecimals);
+
+    // Adjusting the multiplier for easier use later.
+    // “The current exchange rate as an unsigned integer, scaled by
+    //   1 * 10 ^ (18 - 8 + Underlying Token Decimals)” - https://compound.finance/docs/ctokens#exchange-rate
+    //   Simplified to 10^(10 + Underlying Token Decimals).
+    const exchangeDecimals = new BigNumber(10).pow(10);
+    tokens[tokenAddress].exchangeRateMult = exchangeRate
+      .dividedBy(exchangeDecimals).dividedBy(tokens[tokenAddress].tokenDecimalsMult);
```
echoAE, line 88: To make this consistent with the variable naming elsewhere:
```suggestion
    const exchangeDecimalsMult = new BigNumber(10).pow(10);
    tokens[tokenAddress].exchangeRateMult = exchangeRate
      .dividedBy(exchangeDecimalsMult).dividedBy(tokens[tokenAddress].tokenDecimalsMult);
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,389 @@
...
+    const pages = [...Array(maxPages)].map((_, i) => 1 + i);
+    await Promise.all(pages.map(async (page) => {
+      const currentRequest = buildJsonRequest(
+        maximumHealth, minimumBorrowInETH, page, pageIncrement,
+      );
+      const apiResults = await callCompoundAPI(currentRequest);
+      apiResults.accounts.forEach((currentAccount) => {
+        foundAccounts.push(currentAccount);
+      });
+    }));
```
echoAE, line 159: Instead of pushing the individual elements of the `apiResults.accounts` Array into the `foundAccounts` Array, you should be able to wait for all of the promises to settle, which should return an Array of Arrays, then use `.flat()` to create `foundAccounts`.
```suggestion
    // Query each page and add accounts. Starting at 1 and including maxPages
    // Shorthand Range() function. ( Ex: 5 => [1,2,3,4,5] )
    const pages = [...Array(maxPages)].map((_, i) => 1 + i);
    const foundAccounts = await Promise.all(pages.map(async (page) => {
      const currentRequest = buildJsonRequest(
        maximumHealth, minimumBorrowInETH, page, pageIncrement,
      );
      const apiResults = await callCompoundAPI(currentRequest);
      return apiResults.accounts;
    })).flat();
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,389 @@
...
+    // Initialize accounts. New accounts will get updated in the block section
+    data.newAccounts.forEach((newAccount) => { accounts[newAccount.toLowerCase()] = {}; });
+    data.totalNewAccounts += data.newAccounts.length;
+    data.newAccounts = [];
+
+    // Update Balances on zero health accounts
+    const filteredAccounts = [];
+    Object.keys(accounts).forEach((currentAccount) => {
+      if (accounts[currentAccount].health == null
+        || accounts[currentAccount].health === 0) {
```
echoAE, line 244:
```suggestion
    Object.entries(accounts).forEach(([currentAccount, entry]) => {
      if (entry.health == null || entry.health === 0) {
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,389 @@
...
+
+    // Calculate health on all accounts
+    Object.keys(accounts).forEach((account) => {
+      let borrowBalance = new BigNumber(0);
+      let supplyBalance = new BigNumber(0);
+      Object.keys(tokens).forEach((token) => {
+        if (supply[token][account]) {
+          // Supply balances are stored in cTokens and need to be multiplied by the
+          //   exchange rate, price and the collateral factor to determine the ETH value
+          //   of the collateral.
```
echoAE, line 318:
```suggestion
          //   of the collateral.
          // Ref: https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/Comptroller.sol#L751
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,389 @@
...
+            .multipliedBy(tokens[token].exchangeRateMult)
+            .multipliedBy(tokens[token].price)
+            .multipliedBy(tokens[token].collateralMult));
+        }
+        if (borrow[token][account]) {
+          // Only need to multiply by the price.
+          borrowBalance = borrowBalance.plus(borrow[token][account]
+            .multipliedBy(tokens[token].price));
+        }
+      });
```
echoAE, line 329:
```suggestion
      Object.entries(tokens).forEach(([token, entry]) => {
        const {
          exchangeRateMult,
          price,
          collateralMult,
        } = entry;
        
        // Ref: https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/Comptroller.sol#L751
        const tokensToDenom = exchangeRateMult.times(price.times(collateralMult));
        
        if (supply[token][account]) {
          // Supply balances are stored in cTokens and need to be multiplied by the
          //   exchange rate, price and the collateral factor to determine the ETH value
          //   of the collateral.
          // Ref: https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/Comptroller.sol#L754
          supplyBalance = supplyBalance.plus(supply[token][account].times(tokensToDenom));
        }
        
        // Ref: https://github.com/compound-finance/compound-protocol/blob/3affca87636eecd901eb43f81a4813186393905d/contracts/Comptroller.sol#L757
        if (borrow[token][account]) {
          // Only need to multiply by the price.
          borrowBalance = borrowBalance.plus(borrow[token][account].times(price));
        }
      });
```

liquidation-monitor/src/agent.js
Nomad-AE: Implemented in a5e28621d461abcaac565530d37c2979cd9518ba

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,389 @@
...
+        if (
+          token.supply_balance_underlying !== undefined
+          && token.supply_balance_underlying.value !== 0
+        ) {
+          data.supply[token.address][account.address] = new BigNumber(
+            token.supply_balance_underlying.value,
+          ).dividedBy(data.tokens[token.address].exchangeRateMult);
+        }
+      }); // end token loop
+    }); // end account loop
```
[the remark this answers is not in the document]
Nomad-AE: Agreed, this is better style.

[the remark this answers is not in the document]
Nomad-AE: Implemented in commit ce859a7125ada79a96365c269b1c7350564ede7b

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,389 @@
...
+
+      // Adjust for native decimals
+      tokens[currentToken].price = new BigNumber(price.toString()).dividedBy(oneInchMult);
+
+      // Update the Collateral Factor
+      const market = await comptrollerContract.markets(currentToken);
+      // Per Compound: https://compound.finance/docs/comptroller#collateral-factor
+      //   "collateralFactorMantissa, scaled by 1e18, is multiplied by a supply balance to determine
+      //    how much value can be borrowed"
+      tokens[currentToken].collateralMult = new BigNumber(market[1].toString())
```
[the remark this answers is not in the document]
Nomad-AE: I attempted it earlier and had some issues. But your code does indeed work. I do like this pattern better.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,401 @@
+const {
+  Finding,
+  FindingSeverity,
+  FindingType,
+  ethers,
+  getEthersProvider,
+} = require('forta-agent');
+const BigNumber = require('bignumber.js');
+const config = require('../agent-config.json');
```
echoAE, line 9:
```suggestion
const config = require('../bot-config.json');
```

liquidation-monitor/agent-config.json
```diff
@@ -0,0 +1,21 @@
+{
```
echoAE, line 1: Rename this file `bot-config.json`

liquidation-monitor/src/agent.spec.js
```diff
@@ -0,0 +1,309 @@
...
+  ethers: {
+    ...jest.requireActual('ethers'),
+    Contract: jest.fn().mockReturnValue(mockContract),
+  },
+}));
+
+const {
+  Finding, FindingType, FindingSeverity,
+} = require('forta-agent');
+const config = require('../agent-config.json');
```
echoAE, line 42:
```suggestion
const config = require('../bot-config.json');
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,401 @@
...
+      }));
+    }));
+
+    // Update all token prices via 1inch
+    await Promise.all(Object.entries(tokens).map(async ([currentToken, entry]) => {
+      const price = await oneInchContract.getRateToEth(entry.underlying, 0);
+
+      // 1inch's getRateToEth is scaled to 1e18 but is also affected by the underlying token decimal
+      //   using the calculation: 10^18 / 10^(tokenDecimal) - https://docs.1inch.io/docs/spot-price-aggregator/examples
+      //   Combining the two scaling factors: 10^(18 + 18 - tokenDecimal) => 10^(36 - tokenDecimal)
```
echoAE, line 299: This is a bit of a verbose suggestion.  I'm trying to make it abundantly clear what the returned value from the 1inch contract is and the different scalings that must be applied to arrive at the price.

Feel free to simplify the comment and reduce the verbosity, as necessary.
```suggestion
      // 1inch's getRateToEth is scaled to 1e18 but is also affected by the underlying token decimal
      //   getRateToEth = (amount of Token / amount of Ether) * 10^18
      // To account for token decimals, Ether decimals, and the scaling, we need to perform the following
      //   price = getRateToEth * (10^tokenDecimals) * (1 / 10^etherDecimals) * (1 / 10^18)
      // Because the number of Ether decimals is fixed at 18, we can simplify the expression
      //   price = getRateToEth * (10^tokenDecimals / 10^36)
      //     or
      //   price = getRateToEth * 10^(tokenDecimals - 36)
      //     or
      //   price = getRateToEth / (10^(36 - tokenDecimals))
      // Ref: https://docs.1inch.io/docs/spot-price-aggregator/examples
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,401 @@
...
+      }
+    });
+
+    // Check the Comptroller contract for actual liquidity if the health is below x
+    const lowHealthAccounts = [];
+    Object.keys(accounts).forEach((currentAccount) => {
+      if (accounts[currentAccount].health.isLessThan(data.lowHealthThreshold)) {
+        lowHealthAccounts.push(currentAccount);
+      }
+    });
```
echoAE, line 359: The variable names make this exceed the 100 character limit, but `filter` should do what you want.
```suggestion
    const lowHealthAccounts = Object.keys(accounts).filter((key) => 
      accounts[key].health.lt(data.lowHealthThreshold)
    );
```

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,401 @@
...
+    await Promise.all(lowHealthAccounts.map(async (currentAccount) => {
+      const liquidity = await comptrollerContract.getAccountLiquidity(currentAccount);
+      const shortfallUSD = new BigNumber(ethers.utils.formatEther(liquidity[2]).toString());
```
echoAE, line 362: My guess is that the returned value in `liquidity[2]` will be represented as an `ethers.BigNumber`.  If that's true, then there _should_ be a `toString()` method available on it.
```suggestion
      const shortfallUSD = new BigNumber(liquidity[2].toString());
```

liquidation-monitor/src/agent.js
Nomad-AE: Updated in commit e08bc045b2f90f2087c0808b37ef836152ae46b4

liquidation-monitor/agent-config.json
Nomad-AE: Updated in commit e08bc045b2f90f2087c0808b37ef836152ae46b4

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,410 @@
...
+
+      // Health factor affects the liquidatable amount. Ex: Shortfall of $50 with a Health factor
+      // of 0.50 means that only $25 can be successfully liquidated. ( $25 supplied / $50 borrowed )
```
echoAE, line 374: Reminder to update this with a real-world example pulled from the Compound API.

liquidation-monitor/src/agent.js
```diff
@@ -0,0 +1,410 @@
...
+          data.alert.severity,
+          currentAccount,
+          liquidationAmount.dp(2),
+          shortfallUSD.dp(2),
+          accounts[currentAccount].health.dp(2),
+        );
+        findings.push(newFinding);
+      }
+      // Zero out the health on the low accounts so they may be re-scanned. (optional)
+      // accounts[currentAccount] = {};
```
echoAE, line 394: I can't quite determine whether this should be uncommented or not.  It seems like leaving it commented would result in not re-checking the token balances for accounts with a (currently) low health.  I will give more thought to uncommenting this, but perhaps you have a leaning one way or another.

Nomad-AE: In the `lowHealthThreshold` setting, accounts under 1.03 health ratio will be checked for liquidity in the Comptroller contract. Uncommenting this line would also cause the script to re-check on chain for all account balances. Since the script is also watching for negative health actions such as borrows and withdraw collateral. Then it is not possible to for the balances to change undetected. 

The one reason for the rescan, is that the bot doesn't account for borrowing interest, so the longer the bot runs, the farther the data will skew. Accounts near the threshold can be rescanned for the most accurate balances. Or this can be done on some/all accounts every x blocks.

I'm leaning against using this rescan line. Current workarounds are to increase the threshold or restart the bot.

liquidation-monitor/src/agent.js
Nomad-AE: `formatEther` was a quick call to convert 1e18 wei to 1 Ether. If I just do `toString()` then I'll additionally need to divide by 1e18 to get decimal numbers for outputs.

liquidation-monitor/src/agent.spec.js
```diff
@@ -48,14 +101,31 @@ jest.mock('forta-agent', () => ({
...
+mockBtcDecimals = new ethers.BigNumber.from(mockBtcDecimals);
+mockEthDecimals = new ethers.BigNumber.from(mockEthDecimals);
+mockUsdcDecimals = new ethers.BigNumber.from(mockUsdcDecimals);
+mockBtcCollateralFactor = new ethers.BigNumber.from(mockBtcCollateralFactor);
+mockEthCollateralFactor = new ethers.BigNumber.from(mockEthCollateralFactor);
+mockUsdcCollateralFactor = new ethers.BigNumber.from(mockUsdcCollateralFactor);
+mockBtcCTokenRate = new ethers.BigNumber.from(mockBtcCTokenRate);
+mockEthCTokenRate = new ethers.BigNumber.from(mockEthCTokenRate);
+mockUsdcCTokenRate = new ethers.BigNumber.from(mockUsdcCTokenRate);
+const mockEthersZero = new ethers.BigNumber.from(0);
```
echoAE, line 127: It looks like the constructor cannot be called directly, so taking out the `new` keyword seems appropriate:
```suggestion
mockBtcPrice = ethers.BigNumber.from(mockBtcPrice);
mockEthPrice = ethers.BigNumber.from(mockEthPrice);
mockUsdcPrice = ethers.BigNumber.from(mockUsdcPrice);
mockCDecimals = ethers.BigNumber.from(mockCDecimals);
mockBtcDecimals = ethers.BigNumber.from(mockBtcDecimals);
mockEthDecimals = ethers.BigNumber.from(mockEthDecimals);
mockUsdcDecimals = ethers.BigNumber.from(mockUsdcDecimals);
mockBtcCollateralFactor = ethers.BigNumber.from(mockBtcCollateralFactor);
mockEthCollateralFactor = ethers.BigNumber.from(mockEthCollateralFactor);
mockUsdcCollateralFactor = ethers.BigNumber.from(mockUsdcCollateralFactor);
mockBtcCTokenRate = ethers.BigNumber.from(mockBtcCTokenRate);
mockEthCTokenRate = ethers.BigNumber.from(mockEthCTokenRate);
mockUsdcCTokenRate = ethers.BigNumber.from(mockUsdcCTokenRate);
const mockEthersZero = ethers.BigNumber.from(0);
```

Ref: https://docs.ethers.io/v5/api/utils/bignumber/#BigNumber--creating

liquidation-monitor/src/agent.spec.js
```diff
@@ -161,125 +231,400 @@ describe('check agent configuration file', () => {
...
+    getRateToEth,
+    decimals,
+    getAccountSnapshot,
+    symbol,
+    underlying,
+    exchangeRateStored,
+  } = mockContract;
+
+  // Comptroller
+  /* eslint-disable new-cap */
```
echoAE, line 264: It doesn't look like this comment disabling an ESLint warning needs to be here, as there is no constructor being called.
```suggestion
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -161,125 +231,400 @@ describe('check agent configuration file', () => {
...
+  beforeEach(async () => {
+    // Initialize
+    initializeData = {};
+    setDefaultMocks();
+    setVerifyTokenMocks('cBTC', '0x0wbtc', mockBtcCTokenRate, mockBtcDecimals);
+    setVerifyTokenMocks('cETH', '0x0weth', mockEthCTokenRate, mockEthDecimals);
+    await (provideInitialize(initializeData))();
+  });
+
+  it('should use axios 2 times}', async () => {
```
echoAE, line 317:
```suggestion
  it('should use axios 2 times', async () => {
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -161,125 +231,400 @@ describe('check agent configuration file', () => {
...
+    // Check counter from the initialize step.
+    expect(axios.post).toBeCalledTimes(2);
+  });
+
+  it('should use contract calls}', async () => {
```
echoAE, line 322:
```suggestion
  it('should use contract calls', async () => {
```

liquidation-monitor/src/agent.js
Nomad-AE: Updates in commit 17ed2a115215bd20ef5be24f23bd2a6ee856d910

liquidation-monitor/src/agent.js
Nomad-AE: Clarified in commit 17ed2a115215bd20ef5be24f23bd2a6ee856d910

liquidation-monitor/src/agent.js
Nomad-AE: Removed this entry in commit 17ed2a115215bd20ef5be24f23bd2a6ee856d910 . Also updated README to clarify the limitation.

liquidation-monitor/src/agent.spec.js
```diff
@@ -161,125 +231,400 @@ describe('check agent configuration file', () => {
...
+    const expectedBorrow = '0.85';
+    expect(actualSupply).toBe(expectedSupply);
+    expect(actualBorrow).toBe(expectedBorrow);
+  });
+
+  it('should set BTC token data', async () => {
+    // Check BTC stats
+    const actualBtcDecimals = initializeData.tokens['0x0cbtc'].tokenDecimals.toString();
+    const expectedBtcDecimals = mockBtcDecimals.toString();
+    const actualBtcCDecimals = initializeData.tokens['0x0cbtc'].tokenDecimals.toString();
```
echoAE, line 349:
```suggestion
    const actualBtcCDecimals = initializeData.tokens['0x0cbtc'].cTokenDecimals.toString();
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -161,125 +231,400 @@ describe('check agent configuration file', () => {
...
+    const expectedBtcCDecimals = mockCDecimals.toString();
+    expect(actualBtcDecimals).toBe(expectedBtcDecimals);
+    expect(actualBtcCDecimals).toBe(expectedBtcCDecimals);
+  });
+
+  it('should set ETH token data', async () => {
+    // Check ETH stats
+    const actualEthDecimals = initializeData.tokens['0x0ceth'].tokenDecimals.toString();
+    const expectedEthDecimals = mockEthDecimals.toString();
+    const actualEthCDecimals = initializeData.tokens['0x0ceth'].tokenDecimals.toString();
```
echoAE, line 359:
```suggestion
    const actualEthCDecimals = initializeData.tokens['0x0ceth'].cTokenDecimals.toString();
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -161,125 +231,400 @@ describe('check agent configuration file', () => {
...
+function getShortFallUSD(borrowEth, supplyEth) {
+  const scale = 1000;
+
+  // Liquidity calculations using JS BigNumber
+  let shortfall = new BigNumber(borrowEth).minus(supplyEth);
+  if (shortfall.isNegative()) { shortfall = new BigNumber(0); }
+
+  // Scale to 1e18 for mocking and convert to ethers.Bignumber
+  shortfall = shortfall.times(scale).dp(0).toString(); // Scale out and remove decimals
+  const e18Multiplier = new ethers.BigNumber.from(10).pow(18);
```
echoAE, line 396:
```suggestion
  const e18Multiplier = ethers.BigNumber.from(10).pow(18);
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -161,125 +231,400 @@ describe('check agent configuration file', () => {
...
+  shortfall = new ethers.BigNumber.from(shortfall).mul(e18Multiplier).div(scale);
```
echoAE, line 397:
```suggestion
  shortfall = ethers.BigNumber.from(shortfall).mul(e18Multiplier).div(scale);
```

liquidation-monitor/src/agent.spec.js
Nomad-AE: Fixed all instances of this in commit 7717368c93664a46254a77ad97ed46aebd8f3dcf

liquidation-monitor/src/agent.spec.js
Nomad-AE: Fixed all instances of this in commit 7717368c93664a46254a77ad97ed46aebd8f3dcf

liquidation-monitor/src/agent.spec.js
```diff
@@ -0,0 +1,1000 @@
...
+  axios.post.mockClear();
+  symbol.mockClear();
+  underlying.mockClear();
+  exchangeRateStored.mockClear();
+  decimals.mockClear();
+  getAssetsIn.mockClear();
+}
+
+function setVerifyTokenMocks(setSymbol, setUnderlying, setExchange, setDecimals) {
+  // Verify token MockOnce - BTC then ETH
```
echoAE, line 287:
```suggestion
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -0,0 +1,1000 @@
...
+  let shortfall = new BigNumber(borrowEth).minus(supplyEth);
+  if (shortfall.isNegative()) { shortfall = new BigNumber(0); }
+
+  // Scale to 1e18 for mocking and convert to ethers.Bignumber
+  shortfall = shortfall.times(scale).dp(0).toString(); // Scale out and remove decimals
+  const e18Multiplier = ethers.BigNumber.from(10).pow(18);
+  shortfall = ethers.BigNumber.from(shortfall).mul(e18Multiplier).div(scale);
+
+  // Convert to USD price by dividing by the 1inch exchange rate for USDC.
+  //  Also needs to be scaled my the 1inch multiplier.
```
echoAE, line 398:
```suggestion
  //  Also needs to be scaled by the 1inch multiplier.
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -0,0 +1,1000 @@
...
+function mockTransaction(mockAddress, mockTopic, mockData) {
+  const mockLogs = [{
+    address: mockAddress,
+    topics: mockTopic,
+    data: mockData,
+  }];
+
+  // build the mock receipt for mock txEvent, in this case the log event topics will correspond to
+  // create the mock txEvent
+  const mockTxEvent = new TransactionEvent(null, null, null, [], [], [], mockLogs);
```
echoAE, line 811:
```suggestion
  const mockTxEvent = new TransactionEvent(null, null, null, [], {}, null, mockLogs, null);
```

echoAE (approved):
Very minor update to how TransactionEvent constructor is being called, plus a nit for a typographical error.

PR #17, opened 2022-04-29
AE-1306 Update Compound Agents to bots

darian-ae:
This PR changes all references from Compound `Agents` to `Bots`. The `agent.config.js` file has also been renamed to `bot.config.js` and these bots have now been redeployed. There shouldn't be any new features added or breaking changes introduced in this PR, as it was just renaming changes.


ctoken-monitor/publish.log
```diff
@@ -0,0 +1 @@
+Mon, 21 Mar 2022 18:15:44 GMT: successfully pushed image with reference bafybeiesfylfhlogmav6eacywpzidmcl2x6cab6xr5s53rcbtrle26fwce@sha256:c4781b2ba98db27d3873b71cf89fc251e8bdbc6a6d0e390427878ff900dbd0b8
```
echoAE, line 1: We have not been committing the `publish.log` files, so this can be removed from the pull request.

governance/publish.log
```diff
@@ -0,0 +1 @@
+Mon, 21 Mar 2022 18:25:25 GMT: successfully pushed image with reference bafybeicaui2aelvdtqu4kcgbziwfqmzdyl52ox5dvims3f4owjqd23cyzq@sha256:84dc69e61df2f81b7d71ee049991f0a5e1e498fb2e84b53ee7636a7cbe532da1
```
echoAE, line 1: Remove from pull request.

large-borrows-governance/publish.log
```diff
@@ -0,0 +1 @@
+Mon, 21 Mar 2022 18:32:37 GMT: successfully pushed image with reference bafybeih7qkkwpx5hhti42niwnefk7juoh6rcjovsnhltsfm4iyizeltgxy@sha256:8f1e26668eb97ec2f63a91225d973e3a2bef32e22c2e649b6e2b7a98d0184401
```
echoAE, line 1: Remove from pull request.

multisig-transactions-monitor/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "forta-compound-finance-multisig-transaction-agent",
+  "name": "forta-compound-finance-multisig-transaction-bot",
```
echoAE, line 2: This string is used to determine the Bot ID by the `forta-agent` package.  Changing any characters in the string will result in a different Bot ID, so when the Bot is deployed, it will not update the old one but rather deploy a new one.  Unfortunately, in this case it is unavoidable because the word `agent` appears in the string.

What this means is that for this Bot, as well as any other Bots that updates made to the `name` field, we will need to manually disable the old Bot.

multisig-transactions-monitor/package.json
```diff
@@ -1,7 +1,7 @@
...
   "version": "0.0.1",
```
echoAE, line 3: While a trivial change (and technically a new deployment based on the Bot ID), this should be incremented to version `0.0.2`.

oracle-price-monitor/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "forta-ae-oracle-price-monitor",
   "version": "0.0.1",
```
echoAE, line 3: Increment the version to `0.0.2`

low-liquidity-market-attack-monitor/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "forta-ae-comp-market-attack-events",
   "version": "0.0.1",
```
echoAE, line 3: Increment the version to `0.0.2`

asset-monitor/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "forta-ae-comp-asset-monitor",
   "version": "0.0.1",
```
echoAE, line 3: Increment the version to `0.0.2`

ctoken-monitor/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "forta-ae-comp-ctoken-events",
   "version": "0.0.1",
```
echoAE, line 3: Increment version to `0.0.2`

distribution/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "forta-ae-comp-distribution-bugs",
   "version": "0.0.1",
```
echoAE, line 3: Increment version to `0.0.2`

governance/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "forta-ae-comp-governance-events",
   "version": "0.0.3",
```
echoAE, line 3: Increment version to `0.0.4`

echoAE (changes requested):
- We will need to any duplicate Bots, but ONLY AFTER we have updated the Defender Forta Sentinels that monitor them.  We'll have to switch over from the old Bot IDs to the new Bot IDs.
- Update package versions and redeploy.

Also, it looks like there may be a few more instances of the word `agent`:
- large-borrows-governance/package.json, line 7
- governance/src/agent.js, line 5
- gnosis-safe-deploy/src/agent.js, line 12
- gnosis-safe-deploy/scripts/proposeFortaAgent.js (this was merged in last week)
- oracle-price-monitor/src/agent.spec.js, line 86

multisig-transactions-monitor/package.json
darian-ae: just added

Nomad-AE (approved):
PR looks good, I don't see any breaking changes.

PR #18, opened 2022-05-03
AE-1245: Update compound distribution bot

arbitrary-CodeBeholder:
reworked agent to support sane distribution
updated tests
fixed eslint issues
updated readmes


distribution/package.json
```diff
@@ -18,8 +18,8 @@
     "test": "jest"
   },
   "dependencies": {
-    "bignumber.js": "^9.0.1",
-    "forta-agent": "^0.0.39"
+    "bignumber.js": "^9.0.2",
+    "forta-agent": "^0.1.1"
```
arbitrary-sparrow, line 22:
```suggestion
    "forta-agent": "^0.1.3"
```

distribution/src/agent.js
```diff
@@ -75,75 +101,97 @@ function provideInitialize(data) {
...
+    const compoundTokenAddress = compoundToken.address.toLowerCase();
+    const compoundTokenContract = new ethers.Contract(
+      compoundTokenAddress, decimalsAbi, data.provider,
+    );
+    const compoundTokenDecimals = await compoundTokenContract.decimals();
 
-    data.compDecimalsMultiplier = new BigNumber(10).pow(compDecimals);
-    data.compAddress = compAddress;
-  }
+    data.compoundTokenDecimalsMultiplier = new BigNumber(10).pow(compoundTokenDecimals);
```
arbitrary-sparrow, line 110: This operation is mixing `ethers.js` and `bignumber.js` `bignumbers` since the value of `compoundTokenDecimals` came from a contract call. Converting `compoundTokenDecimals` to a string should prevent any inconsistent behavior:
```suggestion
    data.compoundTokenDecimalsMultiplier = new BigNumber(10).pow(compoundTokenDecimals.toString());
```

distribution/src/agent.spec.js
```diff
@@ -37,393 +38,506 @@ function createTransactionEvent(txObject) {
...
+      const amountCompDistributedBN = new BigNumber('300');
+      const prevBlockCompAccruedBN = new BigNumber('100');
+      const accruedToDistributedRatio = amountCompDistributedBN.div(
+        prevBlockCompAccruedBN,
+      ).times(100);
+
+      const expectedFinding = Finding.fromObject({
+        name: `${protocolName} Exceeds Ratio Threshold Distribution Event`,
+        description: `Distributed ${accruedToDistributedRatio.toFixed(0)}% more COMP to ${validReceiverAddress} than expected`,
+        alertId: `${developerAbbreviation}-${protocolAbbreviation}-EXCEEDS-RATIO-THRESHOLD-DISTRIBUTION-EVENT`,
```
arbitrary-sparrow, line 141:
```suggestion
        alertId: `${developerAbbreviation}-${protocolAbbreviation}-EXCEEDS-RATIO-THRESHOLD-DISTRIBUTION-EVENT`,
```

distribution/src/agent.spec.js
```diff
@@ -37,393 +38,506 @@ function createTransactionEvent(txObject) {
...
 
-    const { abiFile, address: ComptrollerAddress } = Comptroller;
-    const { address: CompoundTokenAddress } = CompoundToken;
+    it('returns a proper exceeds sane distribution finding', () => {
+      const amountCompDistributedBN = new BigNumber('20000');
+
+      const expectedFinding = Finding.fromObject({
+        name: `${protocolName} Exceeds Sane Distribution Event`,
+        description: `Distribution of ${amountCompDistributedBN.toString()} COMP to ${validReceiverAddress} exceeds ${maximumSaneDistributionAmount}`,
+        alertId: `${developerAbbreviation}-${protocolAbbreviation}-EXCEEDS-SANE-DISTRIBUTION-EVENT`,
```
arbitrary-sparrow, line 171:
```suggestion
        alertId: `${developerAbbreviation}-${protocolAbbreviation}-EXCEEDS-SANE-DISTRIBUTION-EVENT`,
```

distribution/abi/Comptroller.json
```diff
@@ -2634,22 +2537,6 @@
...
-    ],
-    "name": "updateContributorRewards",
-    "outputs": [],
-    "payable": false,
-    "stateMutability": "nonpayable",
-    "type": "function",
-    "signature": "0x741b2525"
   }
 ]
 }
```
Nomad-AE, line 2542:
```suggestion
}

```
Nit: It's good to have newlines at the end of files for POSIX compliance.

Nomad-AE (commented):
The current calculation of COMP distribution of current block vs previous block might not be sufficient. It doesn't take into account liquidity increases. If a user has $1 deposited and then deposits $1000, their COMP accrual should increase by 1000x and trigger a false positive alert. Perhaps a comparison of COMP accrued vs account liquidity might be a better metric. A lazy way is to use the Comptroller `getAccountLiquidity` to show the value in USD that is being supplied and you could compare that to the value of COMP accrued. ** This method is incomplete because it doesn't account for borrowed Tokens.

distribution/src/agent.js
```diff
@@ -75,75 +101,99 @@ function provideInitialize(data) {
...
+          findings.push(createExceedsRatioThresholdDistributionAlert(
+            protocolName,
+            protocolAbbreviation,
+            developerAbbreviation,
+            accruedToDistributedRatio,
+            transferEvent.args.to,
+            amountCompDistributedBN.toString(),
+            prevBlockCompAccruedBN.toString(),
+          ));
+        }
```
arbitrary-sparrow, line 194: Adding a minimum distributed amount attempts to curtail how often this alert will be fired; however, the primary issue here is that this ratio has no 'ground truth' for what dictates a normal value vs. one that should be alerted on. The ratio doesn't really provide meaningful insight into whether or not COMP is being distributed 'correctly,' so I think we should remove the ratio logic and finding altogether.

arbitrary-CodeBeholder: I created #AE-1325 to fix the heuristic used, I agree that we should remove the ratio. I had thought to do it as part of that ticket but I'm fine with removing it here, I didn't imagine the bot getting re-deployed until the new heuristic is updated though.

README.md
```diff
@@ -13,7 +13,7 @@ to Proposals and Voting.  All alert types and severities are set to Info.
 ### Compound Distribution Monitor
 
 This agent monitors the Compound Finance Comptroller contract for distribution events that exceed a
```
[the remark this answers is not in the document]
arbitrary-sparrow: This sentence in the README still needs 'agent' replaced with 'bot'

arbitrary-sparrow (approved):

distribution/src/agent.js
```diff
@@ -75,75 +77,70 @@ function provideInitialize(data) {
...
+        maliciousAddress = distributionEvent.args.supplier;
+        compIndex = distributionEvent.args.compSupplyIndex;
       }
 
-      // calculate ratio of accrued to distributed COMP
-      const accruedToDistributedRatio = amountCompDistributedBN.div(prevBlockCompAccruedBN).times(100)
-      if (accruedToDistributedRatio.gt(distributionThresholdPercent)) {
-        findings.push(createDistributionAlert(
+      // check against a maximum "sane" distribution amount of COMP
+      //  and alert then even if the previous is zero?
```
echoAE, line 130:
```suggestion
      //  and alert then even if the previous is zero
```

distribution/src/agent.spec.js
```diff
@@ -37,393 +37,253 @@ function createTransactionEvent(txObject) {
...
+    it('contracts key required', () => {
+      expect(typeof (contracts)).toBe('object');
+      expect(contracts).not.toBe({});
+    });
 
-    const { Comptroller, CompoundToken } = contracts;
-    expect(typeof (Comptroller)).toBe('object');
-    expect(Comptroller).not.toBe({});
+    it('contracts key values must be valid', () => {
+      const { Comptroller } = contracts;
```
echoAE, line 94: Any reason for not checking the Compound token address here as well?

echoAE (approved):
A few minor tweaks.

PR #21, opened 2022-05-16
AE-1368: Fix and redeploy multi-sig bot

echoAE:
There were two small issues that needed to be addressed:
- The event name `ProposalThresholdSet` was used in a switch statement for filtering which event the contract emitted.  Unfortunately, the string had a capital `H` in it in the `case` statement, which would have prevented that case from ever being executed (it would have dropped through to the default case).
- There is a function name with the word `Guardian` misspelled.

### DIFF
--- a/multisig-transactions-monitor/package.json
+++ b/multisig-transactions-monitor/package.json
@@ -1,6 +1,6 @@
 {
   "name": "forta-compound-finance-multisig-transaction-bot",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "description": "Forta Bot monitoring transactions for the Compound Finance Community Multisig Contract",
   "chainIds": [
     1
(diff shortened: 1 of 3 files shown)


Nomad-AE (approved):
Changes look good, but I also see a typo at:
https://github.com/arbitraryexecution/compound-monitoring/blob/980b592cd98cba5756f2fb5ce5f8d058c05fa5d0/multisig-transactions-monitor/src/agent.spec.js#L194
Change `in a a tranasction`  to `in a transaction` typo and misspelling.

Also on line: https://github.com/arbitraryexecution/compound-monitoring/blob/980b592cd98cba5756f2fb5ce5f8d058c05fa5d0/multisig-transactions-monitor/src/agent.spec.js#L286
Change `Adimn` to `Admin`

PR #23, opened 2022-05-24
AE-1394: reduced alerts in a 24 hour period

Nomad-AE:
Added some functionality to only report each unique account once every 24 hours. After the 24 hour delay, the alert for that account will be unmuted again.


liquidation-monitor/src/agent.js
```diff
@@ -111,6 +129,18 @@ function provideInitialize(data) {
...
+    let reportingHour = 12; // reset the alerts on this UTC hour. Set to noon UTC
+    reportingHour %= 24; // Modulo to keep it within 24 hours.
+    // Convert JS time in milliseconds to UTC time in seconds.
+    const now = (new Date().getTime()) / 1000;
+    // Calculate midnight of yesterday 00:00:00
+    const oneDay = 86400;
+    const today = now - (now % oneDay); // now minus seconds elapsed since midnight.
+    const oneHour = 3600;
+    data.nextAlertTime = today + (reportingHour * oneHour);
+    data.alertedAccounts = []; // Accounts alerted in the last 24 hours.
```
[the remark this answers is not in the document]
Nomad-AE: It's a feature. In order to to have groups of alerts occur at the same time of day regardless of when the bot is launched, this logic is needed. If I only wanted  alerts to repeat at UTC midnight, then it can be reduced down to ```js
const now = (new Date().getTime()) / 1000;
const oneDay = 86400;
data.nextAlertTime = now - (now % oneDay);```

Having the alert group repeat at noon seemed better than 4:53 pm, or whenever the bot was started.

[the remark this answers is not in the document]
echoAE: I agree with simplifying this, but I suggest doing it in a slightly different way.

I think that Date().getTime() depends upon the system clock, which means it probably shouldn't be relied upon for synchronization.

Because these have access to blockchain data, I think grabbing the current block (yes, I realize it's in the initialization function) might be a better source.

The following suggestion will get the latest block timestamp, subtract whatever fraction of a day has elapsed, then add a day to it for the next alert time.
```
    const latestBlock = data.provider.getBlock('latest');
    data.nextAlertTime = latestBlock.timestamp - (latestBlock.timestamp % 86400) + 86400;
```
Because Scan Nodes may receive and start this Bot at _slightly_ different times, the removal of any fraction of a day should baseline their starting times to a day boundary.  Depending upon when you deploy the Bot, some Scan Nodes may start the Bot just before a 24-hour boundary and some may start it just after, but they should all end up synchronized on the same day boundaries once blocks start getting processed.

For this to work, you would need to only increment the timestamp by exactly 86400 so that all Bots will alert on the same day boundaries.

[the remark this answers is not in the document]
Nomad-AE: The original consideration is that 2 computers might have varying times IE: 1653609699 and 1653609650. But they would both calculate Midnight of May 27th to be 1653609600. That value would be stored in alert time. The following checks would be based on Eth blocktime vs the alert time, which would put all bots in sync. The limitation to this approach is nodes with more than 24 hour time difference would be on different days.

I concur, that blocktime as the starting point would be most accurate. Fixed in commit 31fe567753617fed31cafbda628011d5f686e0ac

liquidation-monitor/src/agent.js
```diff
@@ -25,7 +27,23 @@ function createAlert(
...
   shortfallAmount,
   healthFactor,
+  currentTimestamp,
+  dataObject,
 ) {
+  // Check the alertTimer
+  if (currentTimestamp >= dataObject.nextAlertTime) {
+    // Add 1 day to the nextAlertTime and remove the previous alerted accounts.
+    const oneDay = 86400;
+    dataObject.nextAlertTime += oneDay;
```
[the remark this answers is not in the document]
Nomad-AE: Leaving as is. Using `currentTimeStamp` will skew the time, as it's unlikely that blocks will fall on UTC midnight 00:00:00 every time.

liquidation-monitor/.dockerignore
```diff
@@ -1,3 +1,5 @@
 node_modules
 dist
-forta.config.json
\ No newline at end of file
+forta.config.json
+autotask-config.json
+.env
```
echoAE, line 5:
```suggestion
```

liquidation-monitor/.gitignore
```diff
@@ -1,3 +1,5 @@
 node_modules
 dist
-forta.config.json
\ No newline at end of file
+forta.config.json
+.env
+autotask-config.json
```
echoAE, line 5:
```suggestion
```

liquidation-monitor/package.json
```diff
@@ -23,6 +23,9 @@
   "dependencies": {
     "axios": "^0.27.2",
     "bignumber.js": "^9.0.2",
+    "defender-relay-client": "^1.21.3",
+    "dotenv": "^16.0.1",
+    "ethers": "^5.6.8",
```
echoAE, line 28:
```suggestion
```

liquidation-monitor/src/agent.js
```diff
@@ -111,6 +129,12 @@ function provideInitialize(data) {
     data.newAccounts = []; // New account from transaction events
     data.totalNewAccounts = 0;
 
+    // Calculate the next report time.
+    const latestBlockTimestamp = (await data.provider.getBlock('latest')).timestamp.toString();
+    //   Now minus seconds elapsed since midnight plus 1 day.
+    data.nextAlertTime = latestBlockTimestamp - (latestBlockTimestamp % 86400) + 86400;
```
echoAE, line 135: Oddly enough, this order of operations works, but I think we just got lucky:
```suggestion
    const latestBlockTimestamp = (await data.provider.getBlock('latest')).timestamp;
    //   Now minus seconds elapsed since midnight plus 1 day.
    data.nextAlertTime = latestBlockTimestamp - (latestBlockTimestamp % 86400) + 86400;
```

For example, the results are different if we move the operations into any other order:
```javascript
> const latestBlockTimestamp = 100;
> latestBlockTimestamp - (latestBlockTimestamp % 7) + 7
105
> latestBlockTimestamp.toString() - (latestBlockTimestamp.toString() % 7) + 7  // current ordering
105
> latestBlockTimestamp.toString() + 7 - (latestBlockTimestamp.toString() % 7)  // move the number forward
1005
> 7 + latestBlockTimestamp.toString() - (latestBlockTimestamp.toString() % 7)  // move the number to the front
7098
> 7 - (latestBlockTimestamp.toString() % 7) + latestBlockTimestamp.toString()  // swap the last two terms
'5100'
> - (latestBlockTimestamp.toString() % 7) + 7 + latestBlockTimestamp.toString()  // swap the first two terms
'5100'
> - (latestBlockTimestamp.toString() % 7) + latestBlockTimestamp.toString() + 7  // swap last two terms
'-21007'
```

liquidation-monitor/src/agent.js
```diff
@@ -416,8 +451,12 @@ function provideHandleBlock(data) {
           liquidationAmount.dp(2).toString(),
           shortfallUSD.dp(2).toString(),
           accounts[currentAccount].health.dp(2).toString(),
+          blockEvent.block.timestamp,
+          data,
         );
-        findings.push(newFinding);
+        if (newFinding != null) {
```
echoAE, line 457:
```suggestion
        if (newFinding !== null) {
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -101,7 +101,7 @@ jest.mock('forta-agent', () => ({
 }));
 
 const {
-  ethers, Finding, FindingType, FindingSeverity, TransactionEvent,
+  ethers, Finding, FindingType, FindingSeverity, TransactionEvent, BlockEvent,
```
echoAE, line 104: There is a helper function in the Forta Bot SDK called `createBlockEvent` that effectively does what the function `mockBlock` does.

The main differences are that:
- `createBlockEvent` will use default values for the `type` and `network` arguments if you don't pass anything in, while `mockBlock` will always populate those with `null`
- `createBlockEvent` takes a `block` Object as an argument, whereas `mockBlock` takes a timestamp as an argument and then creates a `block` Object inside it to create the `BlockEvent` Object

You can switch to `createBlockEvent` if it seems convenient, but I can see a case for both.

liquidation-monitor/src/agent.spec.js
```diff
@@ -417,7 +428,11 @@ describe('handleBlock', () => {
 
     setPriceMocks(mockBtcPrice, mockBtcCollateralFactor, mockBtcCTokenRate);
     setPriceMocks(mockEthPrice, mockEthCollateralFactor, mockEthCTokenRate);
-    await (provideHandleBlock(initializeData))();
+
+    // Set the initial alert to 0 seconds and the first block time to 15 seconds
+    initializeData.nextAlertTime = 0;
```
echoAE, line 433:
```suggestion
    // Set the first block time to 15 seconds
```

echoAE: Also, this file should be linted.  It looks like `mockUsdcCollateralFactor` and `mockUsdcCTokenRate` are never used.

liquidation-monitor/src/agent.spec.js
```diff
@@ -796,6 +827,181 @@ describe('handleBlock', () => {
...
+    setPriceMocks(newEthPrice, mockEthCollateralFactor, mockEthCTokenRate);
+    mockContract.getAccountLiquidity.mockResolvedValueOnce(
+      [mockEthersZero, mockEthersZero, shortfallUSD],
+    );
+    blockEvent = mockBlock(45);
+    findings = await handleBlock(blockEvent);
+    // New finding should be empty because it was the same alert within 24 hours.
+    expect(findings).toStrictEqual([]);
+
+    // Next let 24 pass and process a new block
```
echoAE, line 990:
```suggestion
    // Next let 24 hours pass and process a new block
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -836,7 +1045,11 @@ describe('handleTransaction', () => {
 
     setPriceMocks(mockBtcPrice, mockBtcCollateralFactor, mockBtcCTokenRate);
     setPriceMocks(mockEthPrice, mockEthCollateralFactor, mockEthCTokenRate);
-    await (provideHandleBlock(initializeData))();
+
+    // Set the initial alert to 0 seconds and the first block time to 15 seconds
+    initializeData.nextAlertTime = 0;
```
echoAE, line 1050:
```suggestion
    // Set the initial alert to 0 seconds
```

liquidation-monitor/src/agent.spec.js
```diff
@@ -922,7 +1139,11 @@ describe('process newBorrower', () => {
...
```
echoAE, line 1144:
```suggestion
    // Set the initial alert to 0 seconds
```

echoAE (commented):
Remove `.toString()` from the timestamp variable, add a mock and default mock value for the `getBlock` method on the provider, and lint.

liquidation-monitor/src/agent.js
Nomad-AE: For some reason I thinking that the output was in ethers bigNumber and I wanted to get away from that. Although the minus does cast it back to number.  Good catch that cast to string is not needed here.

liquidation-monitor/src/agent.js
Nomad-AE: This is intentional. `x != null` is the same as `x !==null || x !==undefined` I do want to match both of those. `undefined` for when no findings are found and `null` for findings that are filtered. I've made it explicit in commit e6c3b366a04c0684716915aae00171dec08e9b6e

liquidation-monitor/src/agent.spec.js
Nomad-AE: Ok, I can properly mock it out.

Nomad-AE: Done! Addressed in commit eb58f1732afc2ab71e40882b9760a78a50a8a44a

liquidation-monitor/src/agent.spec.js
Nomad-AE: Addressed in commit eb58f1732afc2ab71e40882b9760a78a50a8a44a

liquidation-monitor/src/agent.spec.js
Nomad-AE: Addressed in commit eb58f1732afc2ab71e40882b9760a78a50a8a44a

echoAE (approved):

arbitrary-sparrow (approved):

PR #27, opened 2022-05-31
AE-1411: create sentinel and autotask for liquidatable positions bot

Nomad-AE:
Added a simplified autotask for posting messages to discord.

### DIFF
--- a/liquidation-monitor/README.md
+++ b/liquidation-monitor/README.md
@@ -29,3 +29,17 @@ This bot detects when an account on Compound is able to be liquidated.
   - Severity is configurable in `agent-config.json`
   - Type is configurable in `agent-config.json`
   - Metadata field contains the borrower's address, liquidatable amount and total shortfall amount.
+
+## Autotask
+
+This Autotask sends alerts from Openzeppelin's Defender Forta Sentinels to Compound's Discord channel.
+
+### Testing
+
+The Autotask files contain code to facilitate development and testing of OpenZeppelin Defender Autotasks.
+
+Rather than creating Autotask scripts in the Defender Web App and then waiting for appropriate block+chain events to trigger a Forta Sentinel, this code allows a developer to specify mocked finding alert. Those alerts are then fed directly into the Autotask in the same format that they would have if they were occurring live.
+
+### Autotask Setup
+- Create a .env file that contains the name of your discord webhook and the URL for it:
+  - ex.) `discordUrl = "discord_webhook_url"`
--- a/liquidation-monitor/README.md
+++ b/liquidation-monitor/README.md
@@ -38,7 +38,7 @@ This Autotask sends alerts from Openzeppelin's Defender Forta Sentinels to Compo
 
 The Autotask files contain code to facilitate development and testing of OpenZeppelin Defender Autotasks.
 
-Rather than creating Autotask scripts in the Defender Web App and then waiting for appropriate block+chain events to trigger a Forta Sentinel, this code allows a developer to specify mocked finding alert. Those alerts are then fed directly into the Autotask in the same format that they would have if they were occurring live.
+Rather than creating Autotask scripts in the Defender Web App and then waiting for appropriate blockchain events to trigger a Forta Sentinel, this code allows a developer to specify a mocked finding alert. Those alerts are then fed directly into the Autotask in the same format that they would have if they were occurring live.
 
 ### Autotask Setup
 - Create a .env file that contains the name of your discord webhook and the URL for it:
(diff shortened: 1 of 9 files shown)


liquidation-monitor/README.md
```diff
@@ -29,3 +29,17 @@ This bot detects when an account on Compound is able to be liquidated.
...
+
+## Autotask
+
+This Autotask sends alerts from Openzeppelin's Defender Forta Sentinels to Compound's Discord channel.
+
+### Testing
+
+The Autotask files contain code to facilitate development and testing of OpenZeppelin Defender Autotasks.
+
+Rather than creating Autotask scripts in the Defender Web App and then waiting for appropriate block+chain events to trigger a Forta Sentinel, this code allows a developer to specify mocked finding alert. Those alerts are then fed directly into the Autotask in the same format that they would have if they were occurring live.
```
echoAE, line 41:
```suggestion
Rather than creating Autotask scripts in the Defender Web App and then waiting for appropriate blockchain events to trigger a Forta Sentinel, this code allows a developer to specify a mocked finding alert. Those alerts are then fed directly into the Autotask in the same format that they would have if they were occurring live.
```

liquidation-monitor/src/autotask.js
```diff
@@ -0,0 +1,111 @@
...
+  const { alert } = body;
+  if (alert === undefined) {
+    return {};
+  }
+
+  // ensure that the alert key exists within the body Object
+  const { source } = body;
+  if (source === undefined) {
+    return {};
+  }
```
echoAE, line 80: This was an initial approach we took to avoid flooding people's email addresses with Autotask failures (i.e. to avoid alert fatigue), but thinking back on it, it probably is not the correct approach.  Each of these will silently swallow any potential issues, which seems worse (apologies for leaving the legacy code in place for so long).

Updating each of these conditionals to throw with some indication of the issue is what the new(er) Autotasks are doing.  This example is for an Autotask triggered by a Contract Sentinel:
```javascript
  if (autotaskEvent === undefined) {
    throw new Error('autotaskEvent undefined');
  }

  const { secrets, request } = autotaskEvent;
  if (secrets === undefined) {
    throw new Error('secrets undefined');
  }

  // ensure that there is a DiscordUrl secret
  const { GovernanceDiscordUrl: discordUrl } = secrets;
  if (discordUrl === undefined) {
    throw new Error('GovernanceDiscordUrl undefined');
  }

  // ensure that the request key exists within the autotaskEvent Object
  if (request === undefined) {
    throw new Error('request undefined');
  }

  // ensure that the body key exists within the request Object
  const { body } = request;
  if (body === undefined) {
    throw new Error('body undefined');
  }

  // ensure that the alert key exists within the body Object
  const {
    matchReasons,
    hash: transactionHash,
  } = body;
  if (matchReasons === undefined) {
    throw new Error('matchReasons undefined');
  }
```

In this way we are at least controlling when the error is thrown and what information it is provided.  If we didn't do it this way, we might make it further into the logic and then end up with some other error that may be harder to trace.  I'm also open to suggestions on other approaches besides all of this error throwing.

liquidation-monitor/src/autotask.spec.js
```diff
@@ -0,0 +1,108 @@
...
+  return autotaskEvent;
+}
+
+describe('check autotask', () => {
+  it('Runs autotask against mock data and posts in Discord (manual-check)', async () => {
+    const autotaskEvent = createFortaSentinelEvent(mockFinding, mockBlockHash, mockTxHash);
+
+    // run the autotask on the events
+    await handler(autotaskEvent);
+  });
```
echoAE, line 107: This is a pretty good justification for making the `throw Error` modifications suggested in the Autotask code. If no `.env` file is present, there are no `secrets`, so there is no posting to Discord. However, the logic in the Autotask masks any issue and just returns, so this test succeeds.

We should determine a unified approach to testing all of these Autotasks, as they have very common usages patterns (i.e. Sentinel request in, Discord POST request out), so we will probably be able to develop some reusable mocks that generally cover everything.

echoAE (commented):
Update the Autotask to follow the newer `throw Error` approach rather than the older `return {}` approach.  Update the test script to have a test to check for the presence of a Discord webhook in the secrets. You can then have positive and negative test cases for when a webhook is present versus when it's missing.

liquidation-monitor/src/autotask.js
Nomad-AE: Fixed in latest revision

liquidation-monitor/src/autotask.spec.js
Nomad-AE: Added more tests including URL checking and a negative test

Nomad-AE:
Autotask now throws more errors and checks if the discord URL is valid. Also added a negative tests for the URL check.

darian-ae (approved):

PR #121, opened 2022-06-16
AE-1474: Pin npm package versions

echoAE:
This work removes all carat symbols `^` from package.json files in this repository. The purpose of this revision is to ensure that all development, testing, and deployment versions of our code are running with the exact same dependencies, rather than having small changes in patch versions or minor versions. Although we would like to trust that `semver` is being followed faithfully, we are choosing to be defensive about dependencies.


darian-ae (approved):

asset-monitor/package.json
```diff
@@ -18,19 +18,15 @@
...
   "dependencies": {
-    "axios": "^0.27.2",
-    "bignumber.js": "^9.0.2",
-    "forta-agent": "^0.0.39",
-    "web3": "^1.7.1",
-    "web3-eth": "^1.7.1",
-    "defender-relay-client": "^1.19.0",
-    "dotenv": "^16.0.0"
+    "forta-agent": "0.0.39",
+    "web3": "1.7.1",
```
Nomad-AE, line 22:
```suggestion
```
This one isn't used

Nomad-AE (approved):
Looks good

asset-monitor/package.json
echoAE: Removed.

PR #153, opened 2022-10-21
AE-2016: Cleanup repo workflows

arbitrary-nomad:
- Add Megalinter Passing status badge to all repos
- Add scan on pushes to `main`
- Configure `NPM-Audit` to scan recursively
- Configure `NPM-Test` to scan recursively
- Remove Node12 from workflows https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
- Remove `set-output` calls used in `Create PR Output` https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


.github/workflows/npm-audit.yml
```diff
@@ -6,51 +6,46 @@ on:
...
-        run: npm outdated
+        # Single command, all newlines are interpreted as a space
+        run: >
+          find .
+          ! -path "*/node_modules/*"
+          ! -path "*/additional_path/*"
+          -name "package.json"
+          -execdir pwd \;
+          -execdir npm ci \;
+          -execdir npm outdated \;
```
teasipper-ae, line 40: Does this need to follow the same for as the `-execdir npm audit {} +` command below?  I believe the `{} +` will pass the exit status of the `-execdir` up i.e. if the `-execdir` fails, then the entire `find` command will fail.

```suggestion
          -execdir npm outdated {} +
```

teasipper-ae (approved):

darian-ae (approved):

PR #157, opened 2022-11-14
AE-2119: Address HTTP 429 issues with Contract cToken Autotask (AED-732)

arbitrary-echo:
Updated the code to collect all messages for a single transaction, concatenate them into a single message, then send a POST request to Discord with that single message. This should address the HTTP 429 errors that we were receiving from sending too many requests in too short a time to Discord.

### DIFF
--- a/defender/downloaded/Contract_cToken.js
+++ b/defender/downloaded/Contract_cToken.js
@@ -3,9 +3,24 @@ const axios = require('axios');
 const axiosRetry = require('axios-retry');
 const ethers = require('ethers');
 
+// function to calculate the delay until the next request attempt
+// returns a value specified in milliseconds
+function retryDelayFunc(retryCount) {
+  // 300 seconds total in Autotask execution to perform retries
+  // #   time   attempt
+  // 0 -   0s - initial request
+  // 1 -  40s - first retry (40s delay from initial request)
+  // 2 - 120s - second retry (80s delay from first retry)
+  // 3 - 280s - third retry (160s delay from second retry)
+  // this leaves 20s for the rest of the Autotask to execute, plus
+  // whatever time each request takes
+  const delay = (retryCount**2)*40*1000;
+  return delay;
+}
+
 axiosRetry(axios, {
   retries: 3,
-  retryDelay: axiosRetry.exponentialDelay,
+  retryDelay: retryDelayFunc,
 });
 
 // import the DefenderRelayProvider to interact with its JSON-RPC endpoint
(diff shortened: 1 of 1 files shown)


defender/downloaded/Contract_cToken.js
```diff
@@ -3,9 +3,24 @@ const axios = require('axios');
...
+  // this leaves 20s for the rest of the Autotask to execute, plus
+  // whatever time each request takes
+  const delay = (retryCount**2)*40*1000;
+  return delay;
+}
+
 axiosRetry(axios, {
   retries: 3,
-  retryDelay: axiosRetry.exponentialDelay,
+  retryDelay: retryDelayFunc,
```
darian-ae, line 23: For the custom `retryDelayFunc()`, are the attempted delay backoffs matching the comments provided? 

With the way the function is currently written, would it look something like this:

```
first retry delay = (1**2)*40*1000 = 40 seconds
second retry delay = (2**2)*40*1000 = 160 seconds
third retry delay = (3**2)*40*1000 = 360 seconds
```

defender/downloaded/Contract_cToken.js
```diff
@@ -3,9 +3,24 @@ const axios = require('axios');
...
+function retryDelayFunc(retryCount) {
+  // 300 seconds total in Autotask execution to perform retries
+  // #   time   attempt
+  // 0 -   0s - initial request
+  // 1 -  40s - first retry (40s delay from initial request)
+  // 2 - 120s - second retry (80s delay from first retry)
+  // 3 - 280s - third retry (160s delay from second retry)
+  // this leaves 20s for the rest of the Autotask to execute, plus
+  // whatever time each request takes
+  const delay = (retryCount**2)*40*1000;
```
arbitrary-nomad, line 17:
```suggestion
  const delay = (2**retryCount)*20*1000;
```
@darian-ae I think this is what was intended. It matches the comments and allows all retries to finish before the 300.10 second autotask timeout.

arbitrary-nomad (approved):

defender/downloaded/Contract_cToken.js
arbitrary-echo: You're absolutely right. Fixed!

defender/downloaded/Contract_cToken.js
arbitrary-echo: Yes, correct! Fixed!

darian-ae (approved):

defender/downloaded/Contract_cToken.js
```diff
@@ -3,9 +3,24 @@ const axios = require('axios');
 const axiosRetry = require('axios-retry');
 const ethers = require('ethers');
 
+// function to calculate the delay until the next request attempt
+// returns a value specified in milliseconds
+function retryDelayFunc(retryCount) {
+  // 300 seconds total in Autotask execution to perform retries
+  // #   time   attempt
+  // 0 -   0s - initial request
```
teasipper-ae, line 11: It probably doesn't matter too much, but 2^0 power is 1.  So if 0 is the first retryCount, then the delay won't be 0, it'll be 20s

```
> (2**0)*20*1000;
20000
> (2**1)*20*1000;
40000
```

arbitrary-echo: I was fairly certain that it started at `1`, but I made a script to double check. TL;DR: It starts at `1`.

package.json
```json
{
  "name": "tmp",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "axios": "0.27.2",
    "axios-retry": "^3.3.1",
    "express": "^4.18.2"
  }
}
```

index.js
```javascript
const axios = require('axios');
const axiosRetry = require('axios-retry');
const express = require('express');

axiosRetry(axios, {
  retries: 3,
  retryDelay: delay,
});

function delay(retryCount) {
  console.log(retryCount);
  return retryCount*1000;
}

const app = express();

app.get('/500', (req, res) => {
  res.status(500).send('Internal Server Error');
});

app.listen(3000, () => {
  console.log('Server is running on port 3000');
});

async function main() {
  const url = 'http://localhost:3000/500';
  const response = await axios.get(url);
}

main()
  .then(() => process.exit(0))
  .catch((err) => {
    console.error('Error');
    process.exit(1);
  });
```

Output:
```shell
Server is running on port 3000
1
2
3
Error
```

This took a while because the `axios` maintainers apparently broke something when they released `v1.0.0` and later: https://github.com/axios/axios/issues/5089

PR #161, opened 2022-11-29
AE-1761 (AED-604) Integrate Compound Gasless voting with Defender Relayer

darian-ae:
This PR integrates the Compound Gasless voting [endpoint](https://app.swaggerhub.com/apis-docs/arr00/COMP.vote/1.0#/) to use Defender's relayer. There is an Autotask Webhook endpoint setup that accepts POST requests of voting or delegate voting batched transactions. The Autotask will then use the Defender Relayer to submit those transactions on-chain.

There is also a [gist](https://gist.github.com/darian-ae/85d1859175414d48194f56db39a57b84) to show an example use case.


defender/staged/downloaded/Compound_Gasless_Voting.js
```diff
@@ -0,0 +1,95 @@
...
+    signer,
+  );
+
+  // cast vote or delegate vote on-chain
+  const promises = body.map(async (votes) => {
+    if (votes.delegatee !== undefined) {
+      const {
+        address, delegatee, nonce, expiry, v, r, s,
+      } = votes;
+
```
teasipper-ae, line 61: Should we sanity check these values before spending the gas to make the call?

defender/staged/downloaded/Compound_Gasless_Voting.js
```diff
@@ -0,0 +1,95 @@
...
+      // delegate vote on-chain
+      console.debug(`Delgating votes from ${address} to ${delegatee}`);
+      try {
+        await compTokenContract.delegateBySig(delegatee, nonce, expiry, v, r, s);
+      } catch (error) {
+        console.error(error);
+      }
+    } else if (votes.support !== undefined) {
+      const {
+        address, proposalId, support, v, r, s,
```
teasipper-ae, line 71: Should we sanity check these values before spending the gas to make the call?

defender/staged/downloaded/Compound_Gasless_Voting.js
```diff
@@ -0,0 +1,95 @@
...
+        await governanceContract.castVoteBySig(proposalId, support, v, r, s);
+      } catch (error) {
+        console.error(error);
+      }
+    } else {
+      // error is not thrown to prevent one failed transaction from failing the rest in the batch
+      console.error('Action is not a vote or delgate-vote');
+    }
+  });
+  await Promise.all(promises);
```
teasipper-ae, line 92: What happens to these votes if an error occurs?  Are they lost?  Is the voter notified?

defender/staged/downloaded/Compound_Gasless_Voting.js
darian-ae: I believe the API endpoint does the sanity checks for us https://app.swaggerhub.com/apis-docs/arr00/COMP.vote/1.0#/default/delegatePOST - if you try to send a post request to it with invalid fields it will won't accept it

darian-ae: You can test that it does validate signatures by pasting this into the data field and executing a POST request to the API endpoint:
```
{
  "address": "0x1111111111111111111111111111111111111111",
  "delegatee": "0x2222222222222222222222222222222222222222",
  "nonce": 0,
  "expiry": 10,
  "v": "0x1b",
  "r": "0xda4429a9e8e6b54cb101b2df002039f2879ab4ca0e8fae64134942cb81f3e581",
  "s": "0x3b90a37dc078a82dfc418695b1d4473661aa4d24dd874ac68678894ff44a6b27"
}
```

It should give back an error messaging stating "given address does not match signer address"

darian-ae: Added sanity check

defender/staged/downloaded/Compound_Gasless_Voting.js
darian-ae: added sanity check to make sure address had delegated votes and proposal Id was active

defender/staged/downloaded/Compound_Gasless_Voting.js
darian-ae: As of right now, the transaction will fail and the vote will not go through and the failed transaction message just gets logged out. It would be great to let the voter know, but I'm not too sure the best way to handle this, as we don't have any email addresses or other forms of notification channels from the users

teasipper-ae: Yeah, I assumed the autotask had no recourse other then to throw an error.  I was mostly making sure we're throwing the correct errors.

teasipper-ae (approved):

defender/staged/downloaded/Compound_Gasless_Voting.js
```diff
@@ -0,0 +1,122 @@
...
+        } else if (support === 1) {
+          console.debug(`Address ${address} is casting a vote against proposal ID: ${proposalId}`);
+        } else if (support === 2) {
+          console.debug(`Address ${address} is casting a vote to abstain for proposal ID: ${proposalId}`);
+        }
+        // cast vote on-chain
+        try {
+          await governanceContract.castVoteBySig(proposalId, support, v, r, s);
+        } catch (error) {
+          console.error(error);
```
arbitrary-nomad, line 111: If the `castVoteBySig` was unsuccessful, this should throw an error.

arbitrary-nomad: Since you are processing a batch of requests, it might be better to store the errors and throw them after everything else is processed.

defender/staged/downloaded/Compound_Gasless_Voting.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+      }
+    } else {
+      // error is not thrown to prevent one failed transaction from failing the rest in the batch
+      console.error('Action is not a vote or delgate-vote');
+    }
+  });
+  await Promise.all(promises);
+
```
arbitrary-nomad, line 120: Piggybacking on @teasipper-ae 's error-handling comment. You can use the flag `let results = true` and anytime an error is encountered set the flag to false. Then at this final return statement `return results`. This would notify the webhook consumer that something did not work, and the autotask owner could look at the logs for this run for more details. 

Alternatively, the errors could be appended to the results and returned to the caller.

arbitrary-nomad (approved):
Great job, I have a few minor changes

defender/staged/downloaded/Compound_Gasless_Voting.js
```diff
@@ -0,0 +1,122 @@
...
+  try {
+    relayAddress = await signer.getAddress();
+  } catch (error) {
+    console.error('Relay is not working, check if it is connected : ', error);
+    throw error;
+  }
+
+  const relayBalance = await provider.getBalance(relayAddress);
+  if (relayBalance <= 0) {
+    throw new Error('Insufficient funds for Relay, please deposit funds.')
```
arbitrary-echo, line 53:
```suggestion
    throw new Error('Insufficient funds for Relay, please deposit funds.');
```

defender/staged/downloaded/Compound_Gasless_Voting.js
```diff
@@ -0,0 +1,122 @@
...
+    if (votes.delegatee !== undefined) {
+      const {
+        address, delegatee, nonce, expiry, v, r, s,
+      } = votes;
+
+      // check if address has votes to delegate
+      const delegateCompBalance = await compTokenContract.balanceOf(address);
+      if (delegateCompBalance > 0) {
+        // delegate vote on-chain
+        console.debug(`Delgating votes from ${address} to ${delegatee}`);
```
arbitrary-echo, line 83:
```suggestion
        console.debug(`Delegating votes from ${address} to ${delegatee}`);
```

defender/staged/downloaded/Compound_Gasless_Voting.js
arbitrary-echo: Agreed. You can remove the `try...catch`, replace the `await Promise.all(promises)` with `const results = await Promise.allSettled(promises)`, then check the `status` of each settled Promise.

If any of the Promises is rejected, you could throw an Error in the Autotask with the combined reasons (each rejected Promise will have a reason string), or use `console.error()` with it and return false to the caller, depending upon who you want to notify.

My guess is that you don't want the Autotask owner to receive Defender emails when a caller fails to use the Autotask correctly. You probably just want a record of the failed call in the Defender logs and then an appropriate response returned to the caller.

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled
Example: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#examples

defender/staged/downloaded/Compound_Gasless_Voting.js
arbitrary-echo: I think this suggestion, combined with using `Promise.allSettled()`, is probably the right way to go.

defender/staged/downloaded/Compound_Gasless_Voting.spec.js
```diff
@@ -0,0 +1,264 @@
+const mockContract = {
+  castVoteBySig: jest.fn(),
+  delegateBySig: jest.fn(),
```
arbitrary-echo, line 3:
```suggestion
  delegateBySig: jest.fn(),
  getCurrentVotes: jest.fn(),
  state: jest.fn(),

```

defender/staged/downloaded/Compound_Gasless_Voting.spec.js
```diff
@@ -0,0 +1,264 @@
...
+    expiry,
+    v,
+    r,
+    s,
+  };
+
+  return delegateVoteObject;
+}
+
+describe('check autotask', () => {
```
arbitrary-echo, line 65: The mocked functions from each test are persisting into the next test. For example, when the return value for `mockContract.state` is set to `1` in the first test case, it will still be `1` for the second test case if the mocked function is not reset.
```suggestion
describe('check autotask', () => {
  beforeEach(() => {
    mockContract.getCurrentVotes.mockReset();
    mockContract.state.mockReset();
  });
  
```

defender/staged/downloaded/Compound_Gasless_Voting.spec.js
```diff
@@ -0,0 +1,264 @@
...
+    await handler(autotaskEvent);
+
+    expect(mockContract.castVoteBySig).toBeCalledTimes(0);
+    expect(mockContract.delegateBySig).toBeCalledTimes(0);
+    mockContract.castVoteBySig.mockClear();
+    mockContract.delegateBySig.mockClear();
+  });
+
+  it('does not cast votes if the proposal is not active', async () => {
+    mockContract.getCurrentVotes = jest.fn().mockResolvedValue(0);
```
arbitrary-echo, line 125: This test is currently passing, but not for the right reasons:
  - The return value of `mockContract.state` was set by the previous test case and it will have the same value here unless it is reset with `mockReset()` (see comment adding a `beforeEach`).
  - The return value for `mockContract.getCurrentVotes` is set to `0`, which means that `castVoteBySig` won't be called because there are no delegated votes.

To address these concerns, we'll set the state to a value other than `1` and we'll make the `getCurrentVotes` value something greater than `0`.
```suggestion
  it('does not cast votes if there are delegated votes but the proposal is not active', async () => {
    mockContract.state = jest.fn().mockResolvedValue(2);
    mockContract.getCurrentVotes = jest.fn().mockResolvedValue(100);
```

defender/staged/downloaded/Compound_Gasless_Voting.spec.js
```diff
@@ -0,0 +1,264 @@
...
+    const autotaskEvent = createAutotaskEvent(batchedTxs);
+    await handler(autotaskEvent);
+
+    expect(mockContract.castVoteBySig).toBeCalledTimes(2);
+    expect(mockContract.delegateBySig).toBeCalledTimes(0);
+    mockContract.castVoteBySig.mockClear();
+    mockContract.delegateBySig.mockClear();
+  });
+
+  it('does not cast votes the address has 0 delegated votes', async () => {
```
arbitrary-echo, line 95:
```suggestion
  it('does not cast votes if the address has 0 delegated votes', async () => {
```

arbitrary-echo (commented):
In the test script, the proposal state should be reset before each test and then explicitly set within each test case (when it is expected to be used for a test case).

One test case was passing for a reason other than what was expected.

defender/staged/downloaded/Compound_Gasless_Voting.spec.js
darian-ae: added

defender/staged/downloaded/Compound_Gasless_Voting.spec.js
darian-ae: added

defender/staged/downloaded/Compound_Gasless_Voting.js
darian-ae: add this suggestion along with the `Promise.allSettled()`

defender/staged/downloaded/Compound_Gasless_Voting.js
darian-ae: removed `try catch` and used `Promise.allSettled()` and returned `false` if there is a failed transaction

teasipper-ae (approved):

arbitrary-nomad (approved):
Looks good

PR #167, opened 2023-02-28
AE-2427-AED-773: Updated Datadog Forta Detection Bot Health Autotask

arbitrary-echo:
The Autotask was failing with the following log output:
```shell
2023-02-28T05:11:27.780Z	ERROR	Invoke Error 	{"errorType":"TypeError","errorMessage":"Cannot read properties of undefined (reading 'metrics')","stack":["TypeError: Cannot read properties of undefined (reading 'metrics')","    at parseMetricsResponse (/var/task/index.js:53:57)","    at /var/task/index.js:285:21","    at processTicksAndRejections (node:internal/process/task_queues:96:5)","    at async Promise.all (index 6)","    at async exports.handler (/var/task/index.js:297:19)","    at async Runtime.handler (/opt/nodejs/autotask-wrapper.js:83:20)"]}
```

This was traced to the old line:
```javascript
  const { data: { data: { getAgentMetrics: { chains: [{ metrics }] } } } } = response;
```

It appears that the Forta Explorer API has been changed to return results where the `chains` field is an empty Array. Previously this was not the behavior.

The updated code verifies that the `chains` field is:
- Not `undefined`
- Not an empty Array
- Has an element with a `chain_id` value of 1

If any of these conditions is not true, the function will return an empty Object.


teasipper-ae (approved):

arbitrary-nomad (approved):

defender/datadog_forta_detection_bot_health/autotask-1/index.js
```diff
@@ -52,8 +52,17 @@ function parseAgentInformationResponse(response) {
 }
 
 function parseMetricsResponse(response, currentTimestamp) {
-  // const { data: { data: { getAgentMetrics: { metrics } } } } = response;
-  const { data: { data: { getAgentMetrics: { chains: [{ metrics }] } } } } = response;
+  const chains = response?.data?.data?.getAgentMetrics?.chains;
+  if (chains === undefined || chains.length === 0) {
+    return {};
+  }
+  let metrics = [];
```
ggonzalez94, line 59: I think this could be simplified like this:
`const metrics = chains.find(chain => chain.chain_id === 1)?.metrics || [];`

PR #169, opened 2023-04-10
SOL-216 Account name fetch

morganjweaver:
Replaces account name and proposal title API calls to Compound v2, which is about to be deprecated on the 15th, with calls to Tally, a governance monitoring API.


defender/secret-example.yml
jbass-oz, line 1: Please revert changes to this file so we can keep a list of secrets and their stacks.

jbass-oz (changes requested):
Requested a couple changes and reviewed some of the files.  I have not run the tests yet. 
 I noticed remaining Todos.  I'll review again at a later commit.

arbitrary-echo (commented):
Quick note about the Tally API key present in a script.

defender/forta_governance/testScript.js
```diff
@@ -0,0 +1,53 @@
+const axios = require('axios');
+
+const tallyApiKey = 'f38a52dd428c9893d63862b60ebf94f3c0f1f93b117f35b9d0f48a0dd8fae616';
```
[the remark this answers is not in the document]
morganjweaver: file deleted, key inconsequential.  thanks!

defender/forta_governance/autotask-1/index.js
jbass-oz, line 1: This seems like a good use case for a closure to set an apikey variable outside the scope of the handler function for reuse by other functions.  A getter method could check if a private variable is valid and throw if not.  The apikey would not need to be added as a parameter to the functions.

defender/governance_discord_alert/autotask-1/index.js
jbass-oz, line 1: This seems like a good use case for a closure to set an apikey variable outside the scope of the handler function for reuse by other functions. A getter method could check if a private variable is valid and throw if not. The apikey would not need to be added as a parameter to the functions.

defender/governance_discord_alert/serverless.yml
```diff
@@ -17,14 +17,16 @@ custom:
 
   # Contract Sentinel Settings
   monitored-addresses:
-    - "0xc0Da02939E1441F497fd74F78cE7Decb17B66529"
+   # - "0xc0Da02939E1441F497fd74F78cE7Decb17B66529"
```
jbass-oz, line 20: remove line instead of comment

defender/governance_discord_alert/serverless.yml
```diff
@@ -17,14 +17,16 @@ custom:
...
+    - "0x057b512C67A2E8491788A89E5f2532e8B9977F0f"
   # Possible Defender networks: https://github.com/OpenZeppelin/defender-client/blob/master/packages/base/src/utils/network.ts`
-  monitored-network: mainnet
+  monitored-network: goerli #mainnet
```
jbass-oz, line 23: remove comment

defender/secret-example.yml
jbass-oz, line 1: revert

defender/tests/onChainResources/deployDummyGovernor.ts
jbass-oz, line 1: is the empty file necessary?

jbass-oz (changes requested):
some changes requested.  closure not necessary at this time.  will review again tomorrow

defender/governance_discord_alert/autotask-1/index.js
morganjweaver: Passing the var around is a little messy, isn't it?  Can you point to an example of this in Defender?  I know closures check for existing env context--would it look something like a helper function like:

```
function getTallyApiKey(){
  return secrets[tallyApiKeySecretName];
}
```
Then on init we check the getter for `undefined` and otherwise use when needed?

defender/forta_governance/autotask-1/index.js
morganjweaver: See governance_discord_alert comment (identical to what I'd write here)

defender/tests/onChainResources/deployDummyGovernor.ts
morganjweaver: deleted

PR #173, opened 2023-04-26
SOL-217 Deprecate Liquidation Monitor

morganjweaver:
-Since we're simply deprecating the monitor, I've added a DEPRECATION label in both READMEs, leaving the code in case we want to review it for future use.
- Also removed the run-in-prod command in the package.json to prevent accidental deployment

Q: If it's better to simply delete the whole thing, please let me know!

### DIFF
--- a/liquidation-monitor/package.json
+++ b/liquidation-monitor/package.json
@@ -7,7 +7,6 @@
   ],
   "scripts": {
     "start": "npm run start:dev",
-    "start:prod": "forta-agent run --prod",
     "tx": "forta-agent run --tx",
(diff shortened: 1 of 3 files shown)


liquidation-monitor/package.json
```diff
@@ -7,7 +7,6 @@
   ],
   "scripts": {
     "start": "npm run start:dev",
-    "start:prod": "forta-agent run --prod",
```
jbass-oz, line 10: No need to remove the line from package.json.  Please revert.

jbass-oz (approved):

PR #178, opened 2023-06-07
SOL-311 Add multisig autotask and sentinel to source control

morganjweaver:
Added autotask filter code and YAML .sls to PR


morganjweaver:
tested and ready for merge I think

PR #180, opened 2023-06-07
SOL-312 price oracle source control

morganjweaver:
Added price oracle filter autotask and sentinel


defender/forta_oracle_price/serverless.yml
jbass-oz, line 1: Please add the notification channel in this serverless so that all references can be resolved if not already deployed.  Naming convention used in other recent serverless change should work with stackname.discord.
According to recent updates to serverless and defender, the sentinel trigger type should be supported with the '@openzeppelin/defender-serverless' plugin. Also we should update the serverless to use the latest defender-serverless plugin.  I think I have mine updated but I'm still troubleshooting issues.

jbass-oz (changes requested):
Add notification channel and update plugin

defender/forta_oracle_price/serverless.yml
morganjweaver: So the current discord channel is set to ` ${self:resources.Resources.notifications.discord}`; I think I need to include it in the resources list so that Defender understands the component is a Discord channel with name `'${self:custom.name} Security Alerts'`.  Is that ok?

morganjweaver:
Assuming Discord config is ok, I think it's ready to merge.

jbass-oz (approved):

defender/forta_oracle_price/serverless.yml
jbass-oz: I changed the name of the channel

jbass-oz (approved):

PR #188, opened 2023-06-28
Update serverless.yml

jbass-oz:
please add a notification channel reference with variable from secrets for discord url so not to expose in commit and retest deployment to a test account


morganjweaver (approved):

jbass-oz:
My approval is pending test of deploying multiple serverless files to the same environment to ensure there are no collisions.  However, I'm confident that the stackname will prevent overwriting the notification channel or autotask configuration.

